PR #14, opened 2014-09-02
Introduce bloom.Builder.

jrick:
This change separates the old Filter type into two types for a client
(usually a wallet) and a bitcoin server.  These types are Builder and
Filter (for the client and server, respectively).  Builder is used by
a client to create a filterload message to send to a bitcoin full
node.  When this message has been received, the filter can be loaded
(using bloom.FilterLoad) to create the Filter structure.  Data can not
be matched directly against the Builder, and unlike Filter, it is not
safe for concurrent access.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcutil/14)

### DIFF
--- a/bloom/builder.go
+++ b/bloom/builder.go
@@ -1,0 +1,80 @@
+// Copyright (c) 2014-2015 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package bloom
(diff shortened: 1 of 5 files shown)


jrick:
rebased

stevenroose:
Is this PR still relevant?

PR #27, opened 2015-01-02
Add Amount.ToBTC convenience method.

### LINKED ISSUE #26: Add convenience method to Amount type.
The Amount type provides a `.ToUnit` function which allows conversion to various units such as BTC, mBTC, and kBTC.  This is a useful function and I think it should definitely remain, since it is a bit verbose to type and the the most common conversion is by far to BTC, I think a convenience method named `ToBTC` should be added to it.  This would allow callers that already have a `btcutil.Amount` to simply do:

```Go
amount.ToBTC()
```

Instead of:

```Go
amount.ToUnit(btcutil.AmountBTC)
```

jrick:
Closes #26.

### DIFF
--- a/amount.go
+++ b/amount.go
@@ -89,6 +89,11 @@ func (a Amount) ToUnit(u AmountUnit) float64 {
 	return float64(a) / math.Pow10(int(u+8))
 }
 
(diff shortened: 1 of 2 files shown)


davecgh:
OK

PR #36, opened 2015-01-24
Optimize base58 decoding.

jrick:
This change introduces an autogenerated base58 digit table to remove
the need to find the index of a character in the modified base58
alphabet each time.  Additionally, it removes some unnecessary big
integer allocations to cut down on the GC churn.

Before:
BenchmarkBase58Encode         20          64998995 ns/op           0.08 MB/s
BenchmarkBase58Decode         50          35965928 ns/op           0.19 MB/s

Now:
BenchmarkBase58Encode         20          64644351 ns/op           0.08 MB/s
BenchmarkBase58Decode        200           7914748 ns/op           0.86 MB/s


jrick:
Don't merge this yet. It may panic on non-ASCII input due to indexing outside of the 7-bit character range.  I'll fix by extending the digit table to 256 bytes.

davecgh:
I was just about to post that it could panic on non-ASCII.

davecgh:
Also, this needs license headers.

base58/base58.go
```diff
@@ -21,8 +19,8 @@ func Decode(b string) []byte {
 	j := big.NewInt(1)
 
 	for i := len(b) - 1; i >= 0; i-- {
-		tmp := strings.IndexAny(alphabet, string(b[i]))
-		if tmp == -1 {
+		tmp := b58[b[i]]
+		if tmp == 255 {
 			return []byte("")
 		}
 		idx := big.NewInt(int64(tmp))
```
davecgh: I know this was already there, but while you're in the mode of optimizing, there is no need to create a new big integer each time here.  

```Go

var scratch *big.Int
for i := len(b) - 1; i >= 0; i-- {
	...
	scratch.Mul(j, idx)
	answer.Add(answer, scratch)
	j.Mul(j, bigRadix)
}
```

base58/base58.go
```diff
@@ -21,8 +19,8 @@ func Decode(b string) []byte {
 	j := big.NewInt(1)
 
 	for i := len(b) - 1; i >= 0; i-- {
-		tmp := strings.IndexAny(alphabet, string(b[i]))
-		if tmp == -1 {
+		tmp := b58[b[i]]
```
davecgh: Also, since the table is being autogenerated now, would it make sense to just autogenerate the b58 arrays with `big.Ints` instead of doing a lookup and creating a new big integer each time?

base58/base58.go
jrick: I'd actually like to remove most of the big math in this package, but that's a more invasive change.

base58/base58.go
jrick: Sounds like a good idea. let me try it.

davecgh: Disregard this.  I updated my comment above about the scratch with a way to avoid the new big.Int there too.

jrick:
```
BenchmarkBase58Encode         20          64150584 ns/op           0.08 MB/s
BenchmarkBase58Decode         50          29615028 ns/op           0.23 MB/s
```

base58/base58.go
jrick: When using the scratch int:

```
BenchmarkBase58Encode         20          64476739 ns/op           0.08 MB/s
BenchmarkBase58Decode        200           7755468 ns/op           0.88 MB/s
```

davecgh: Nice.  That's a pretty significant difference.

davecgh:
Might want to update the PR description with the new timing for historical purposes.  Looks good.

PR #41, opened 2015-02-06
Update btcnet path import paths to new location.

davecgh:


jrick:
Travis appears to be failing this one due to the circular dependency between this and the btcd repo.  However, the changes look good and we can push this and the package renames for btcd/txscript at the same time and everything should continue building.

PR #43, opened 2015-03-04
Update golang.org/x/crypto import paths to new location.

davecgh:


jrick:
We should remove the whitespace between import paths, since each is now under "github.com/btcsuite/*"

davecgh:
Done.

PR #47, opened 2015-06-11
Add Common Name to certificate.

jrick:
Some applications fail to parse the certificate if the CN is not set,
even if they (correctly) check SANs before the CN when validating a
hostname.  Even though the CN should be ignored if a matching SAN
hostname was found, we can prevent the parse from failing by also
including the hostname as the CN.

Additionally, switch from maps to slices to prevent DNS names and IP
addresses from being reordered when added to the certificate template.


jrick:
Relevant RFC: https://tools.ietf.org/html/rfc612

The usage of the Common Name is not recommended, but it is ok to still include it for backwards compatibility.

jrick:
Travis tests failed for an unrelated problem that is fixed in #48.  Will rebase after that is in.

davecgh:
OK after rebase/squash to single commit.

PR #50, opened 2015-07-20
Improve godoc. Add testable examples.

wallclockbuilder:


davecgh:
Thanks for the PR.  It's always good to have runnable/testable examples.

Would you mind putting the example in an `example_test.go` file for consistency with the rest of the code base?  See base58/bloom/hash160/hdkeychain.

jrick:
Since all of the examples print either a correct result or the conversion error, I would do:

```Go
amt, err := btcutil.NewAmount(...)
if err != nil {
        fmt.Println(err)
} else {
        fmt.Println(amt)
}
```

This removes the need to return early on error.

You can also reuse the same error variable for each call.

davecgh:
I'd rather it return early myself.  It's more consistent with the standard way of handling errors and there is really not much point in printing multiple errors.

I agree that `err` should be reused instead of separate `err1`, `err2`, `err3`, etc.

davecgh:
Also, once you've moved it into `example_test.go`, please squash everything down into a single commit.

jrick:
early returns make no sense if you want to show testable examples for each of the error cases.

wallclockbuilder:
Yeah two ways:
Show testable examples for each of the error cases OR
Show standard production grade code samples on the godoc.
I suspect most people who find the godocs helpful appreciate code that demonstrates a standard way of doing things in production mode.

wallclockbuilder:
Whats breaking the build. I wonder.

jrick:
```
./example_test.go:10: ExampleNewAmount redeclared in this block
	previous declaration at ./amount_test.go:114
```

davecgh:
The example is in both files.  It only should be in `example_test.go`.

wallclockbuilder:
:-1: For me. 
Thanks.

davecgh:
Thanks for the update.  Please squash to a single commit and then it's OK from me.

wallclockbuilder:
Squashed.

wallclockbuilder:
Status check. @davecgh Is this acceptable now?

davecgh:
Yep.  Looks good.  Thanks.  I'll merge it in a second.

OK.

PR #51, opened 2015-07-21
Add testable example for UnitConversions

wallclockbuilder:


const.go
```diff
@@ -13,4 +13,8 @@ const (
 
 	// MaxSatoshi is the maximum transaction amount allowed in satoshi.
 	MaxSatoshi = 21e6 * SatoshiPerBitcoin
+
+	// Satoshi is the base bitcoin monetary unit.
+	// A single amount is equal to 1e-8 of a bitcoin.
```
davecgh: Constant definitions show up in the godoc documentation, so I'm not sure the second comment line adds anything.  It doesn't hurt to have it either I suppose, but I don't think it's providing any additional information.

Also, what do you think of naming the constant `BitcoinPerSatoshi`.  It's a little more consistent with the other constants and I think less confusing in that it conveys it's a multiplier.  To me, the name `Satoshi` and and of itself is the value `1` by virtue of it being the base unit.

jrick:
I'm going to be 'that guy' and put my foot down on the floating point equivalency checks.  While it's certainly possible to do that when everything is a compile time constant, someone will look at this example and then start using it on runtime values.

All comparisons between amounts should be done using the Amount type, not floats.

It's ok to compare the formatted strings however. Those should always be formatted exactly the same.

const.go
jrick: If we want a satoshi constant it should by typed (Amount) and have the value 1.

davecgh:
Agreed about the floating point equality checks.  That is a no-no.  Comparing the formatted strings is fine as well as the amount types themselves.

wallclockbuilder:
Good point. We don't want people performing floating point equality checks on runtime values.

wallclockbuilder:
Instead of `amount.ToUnit(AmountMilliBTC) == 444333222.111`
remove floating point equality checks or replace with:
`amount.ToUnit(AmountMilliBTC) == NewAmount(444333222.111 * Satoshi).ToUnit(AmountMilliBTC)`

wallclockbuilder:
Rather  `amount.ToUnit(AmountMilliBTC) == NewAmount(444333222.111 * SatoshiPerMilliBTC).ToUnit(AmountMilliBTC)`

jrick:
Those are still floats.

const.go
wallclockbuilder: @davecgh Renaming it BitcoinPerSatoshi will be consistent with the rest of the code. Makes sense.
@jrick The satoshi constant is a workaround for the quirky behavior of NewAmount(). Conventionally, all amounts are specified int the base unit; Satoshi. Other units are derivative of that. NewAmount() however expects its monetary value specified in BTC. This is why the Satoshi=1e-8 workaround exists. 
Once NewAmount() takes in Satoshi as default, the `const Satosh = 1` will be true and no more needed.

wallclockbuilder:
What I mean is `amount == NewAmount(444333222.111 * BitcoinPerMilliBTC)`

const.go
jrick: The workaround isn't needed in the first place since you can just type convert integer satoshi amounts to Amount.

```Go
var satoshis int64 = 1.23e8
var amount = btcutil.Amount(satoshis)
fmt.Println(amount == 1.23e8) // true
```

davecgh: In the case of the proposed `example_test.go` code:

```Go
	amount := Amount(4443332221110)
	fmt.Println("Satoshi to kBTC:", amount.Format(AmountKiloBTC))

	// Output:
	// Satoshi to kBTC: 444.333222111
```

davecgh: At any rate I agree there really is no need for the constant.

jrick: If there's confusion between NewAmount and type converting btcutil.Amount, let's rename NewAmount to something more specific (AmountFromBitcoin perhaps), not remove the ability to safely create satoshi integer amounts from floating points (which is necessary for the stupid wallet RPC api).

davecgh:
Also, since this (and the last PR really) are intended to be example code that callers will be looking at, I'd suggest not using a . import since that is not how callers will typically be accessing the functions.

wallclockbuilder:
OK. Will remove the . import

const.go
wallclockbuilder: The more descriptive the name. The less vague. The better.

example_test.go
```diff
@@ -41,3 +41,20 @@ func ExampleNewAmount() {
 	// 0 BTC
 	// invalid bitcoin amount
 }
+
+func ExampleAmount_unitConversions() {
+	amount := btcutil.Amount(44433322211100)
+
+	fmt.Println("Satoshi to kBTC:", amount.Format(btcutil.AmountKiloBTC))
+	fmt.Println("Satoshi to BTC:", amount.String())
```
davecgh: Not that it matters a lot, but no need to call `.String()` here. `btcutil.Amount` is a `Stringer`, so `fmt.Println` will automaticaly invoke it:
```Go
fmt.Println("Satoshi to BTC:", amount)
```

wallclockbuilder: Good point. Thanks. Will update that.

davecgh:
This looks good to me now.  Please squash it back to a single commit.  @jrick This OK for you now?

jrick:
sure

wallclockbuilder:
Theres a confusion between NewAmount() and Amount. 
I'll rename NewAmount() to something more specific.
The name @jrick suggested works: AmountFromBitcoin()
Is that ok?

wallclockbuilder:
```go
        amountNaN, err := btcutil.NewAmount(math.NaN()) //returns err
```
Also ```go
amountNaN = btcutil.NewAmount(math.NaN()) //returns -92233720368.54776 BTC
```
I will add a NewAmountFromSatoshi() that actually takes in the base BTC and returns `err` for invalid inputs.

Your thoughts please.

**Comment by wallclockbuilder** [comment]
```go
        amountNaN, err := btcutil.NewAmount(math.NaN()) //returns err
```
Also ```go
amountNaN = btcutil.NewAmount(math.NaN()) //returns -92233720368.54776 BTC
```
I will add a NewAmountFromSatoshi() that actually takes in the base BTC and returns `err` for invalid inputs.

Your thoughts please.

davecgh:
I'd prefer any name changes to be in a separate PR.  Particularly `NewAmount` since it will need to be accompanied by several other PRs to other repositories which make use of it.

davecgh:
As far as the NaN thing, I honestly don't see much point to it.  Whenever you cast a value, you are explicitly telling the compiler that you know what you're doing.  `math.NaN()` returns a float64, not an int64.

Any amount of Satoshi will always be int64s under the hood.  Even an Amount is an int64, so there really is no legitimate way to end up in a situation you've illustrated.  Your example code is essentially doing int64(someFloat), so it's doing exactly what you told it to and truncating it!

wallclockbuilder:
In the case of NewAmount(), Amount is only invalid if it cannot be represented as an integer type. This is not a problem with Amount() as its always represented as an `int64` value.

`type Amount int64` is perfect as it is. The example in the godoc, will make the intended use obvious for sleep deficient minds like me. 

All thats left is the rename of NewAmount for clarity and avoidance of vagueness.(I'll do that in a separate PR as recommended.)

jrick:
I don't want unit tests for whether casting actually works.  That is provided by the language and it's reasonable to assume it works as intended.  At the assembly level both an Amount and int64 are exactly the same thing and converting between them is a noop.

If adding tests were free I wouldn't particularly care, but tests are just as much an important part of the code and they do need maintenance.  In this case they add no real value and are just extra code to deal with (both for humans and the testing infrastructure).

I'm not opposed to tests, but make them count.

jrick:
Also, the example is now using casting as well, and that is fine (because it is an example of usage), but as the name implies, testable examples are tests and they do run during 'go test'.  Testing that again with table driven tests is unneeded.

wallclockbuilder:
Point well taken. The examples are fine. 
Wrote the test as documentation, and then later on added the testable examples to make the tests visible in on the godoc page. Its now redundant. I've taken it out.

Any other thoughts? Is it a go now?

jrick:
Looks good. Please squash into a single commit.

wallclockbuilder:
Squashed.

davecgh:
OK

PR #53, opened 2015-07-25
Rename NewAmount

wallclockbuilder:
By convention NewAmount is the default constructor for the type
‘Amount’. In this case however thats not the case. I have thus renamed
NewAmount to NewAmountFromBTC(suggested by @jrick ) to make that clearer.

### DIFF
--- a/amount.go
+++ b/amount.go
@@ -68,6 +68,11 @@ func round(f float64) Amount {
 // some value in bitcoin.  NewAmount errors if f is NaN or +-Infinity, but
 // does not check that the amount is within the total amount of bitcoin
 // producible as f may not refer to an amount at a single moment in time.
+//
+// NewAmount is for specifically for converting BTC to Satoshi.
+// For creating a new Amount with an int64 value which denotes a quantity of Satoshi,
+// do a simple type conversion from type int64 to Amount.
+// See GoDoc for example: http://godoc.org/github.com/btcsuite/btcutil#example-Amount
 func NewAmount(f float64) (Amount, error) {
 	// The amount is only considered invalid if it cannot be represented
 	// as an integer type.  This may happen if f is NaN or +-Infinity.


wallclockbuilder:
I further renamed simplified `NewAmountFromBTC` to `ParseBTC`

wallclockbuilder:
The ultimate simplicity is:
```go
oneBTC, err := Amount.BTC(1)
hundredMillionSatoshi = Amount.Satoshi(1e8)
oneBTC == hundredMillionSatoshi
```

wallclockbuilder:
The simple version is succint and very readable. It offers clarity without ambiguity.

wallclockbuilder:
The time(http://godoc.org/time) package uses a similar pattern:
```go
time.Date() // equivalent to NewTimeFromDate()
time.Parse() // equivalent to NewTimeFromString()
time.Unix() // equivalent to NewTimeFromUnixTimeStamp()
```

wallclockbuilder:
I noticed `Type Block`(and others eg: `type Address`, `type AddressScriptHash`) has the NewTypeFromInput constructors. 
```go
/// type Block constructors
func NewBlock(msgBlock *wire.MsgBlock) *Block {...}
func NewBlockFromBlockAndBytes(msgBlock *wire.MsgBlock, serializedBlock []byte) *Block {...}
func NewBlockFromBytes(serializedBlock []byte) (*Block, error) {...}
func NewBlockFromReader(r io.Reader) (*Block, error) {...}

// type AddressScriptHash constructors
func NewAddressScriptHash(serializedScript []byte, net *chaincfg.Params) (*AddressScriptHash, error)
func NewAddressScriptHashFromHash(scriptHash []byte, net *chaincfg.Params) (*AddressScriptHash, 

// type Tx constructors
func NewTx(msgTx *wire.MsgTx) *Tx {...}
func NewTxFromBytes(serializedTx []byte) (*Tx, error) {...}
func NewTxFromReader(r io.Reader) (*Tx, error) {...}
```
Maybe Amount should be consistent with this package-wide pattern.
This is predictable, thus not confusing.

wallclockbuilder:
```go
oneBTC, _ := btcutil.NewAmountFromBTC(1)
hundredMillionSatoshi := btcutil.NewAmountFromSatoshi(1e8)
fmt.Println(oneBTC == hundredMillionSatoshi)
```

wallclockbuilder:
Existing users of the previous `btcutil.NewAmount()` will update to `btcutil.NewAmountFromBTC()`.
New users who expect `btcutil.NewAmount()` to receive Satoshi inputs can use `btcutil.NewAmountFromBTC()`
Win. Win.

davecgh:
Alright, so there are several things going on.

Let me start by saying that personally I don't think it's worth changing because I don't taking changing public APIs for which there are already consumers lightly unless there is a really good reason beyond a name.  That said, let me address the specific points:

There is a fundamental difference between `Amount`s and the other things you've listed (`NewBlock...`, `NewAddress...`, `NewTx...`).  Specifically, all of those other types, including the `time.Time`, are a complex type (structs) which require some form of initialization and or error handling with regards to their creation.

An `Amount`, on the other hand, is simply an int64 with convenience functions on it for doing pretty printing and math conversions.  There are no complex structure creations nor error conditions to handle for "creating" an amount from an an int64 (satoshi).  On the other hand, there are error conditions to be handled when converting from BTC (a non-integral type), hence the constructor function for it.  This pattern is typical in Go where you create a new type which is defined in terms of another one and add some additional functions to it.  Then, a zero overhead cast is performed gain access to the specialized functions.

I prefer not Introducing the additional overhead of a function call for something that is nothing more than a simple type cast.

jrick:
Here are all the consumers of btcutil.NewAmount on github:

https://github.com/search?l=go&q=btcutil.NewAmount&type=Code&utf8=%E2%9C%93

and there is likely more code using this api either not on github or in private repos.

I'm not too thrilled with the idea of breaking all of this code.

A documentation change is sufficient for this case.  I never expected this to be a problem (the documentation already does say it converts from a floating point value representing some number of bitcoin), and it appears that many consumers are using the api as it was intended.  If anything, as @davecgh mentioned, we can just add to the comment that type converting (casting) is the correct method of converting integer amounts of satoshi to Amount.

wallclockbuilder:
I agree: changing a public API for which there is over 30 consumers should not be done lightly.
Breaking all their code? Thats a bad idea. Very bad.

However leaving it as it is doesn't address the fact that NewAmount() takes in BTC amounts instead of Satoshi amounts as conventionally expected.

Your suggestion is good: lets keep NewAmount() as it is. Add in an extra line of documentation which calls out the fact that its specifically for creating `Amount` s from BTC. The advise users to perform a simple type cast to Amount when they need to construct an amount from Satoshi.

How about just adding a line to NewAmount() telling them to use NewAmountFromBTC instead.
That should push use towards the appropriate function without breaking existing codebases.

This is what I'm thinking:
newamountupdate

wallclockbuilder:
```go
// NewAmount creates an Amount from a floating point value representing
// some value in bitcoin.  NewAmount errors if f is NaN or +-Infinity, but
// does not check that the amount is within the total amount of bitcoin
// producible as f may not refer to an amount at a single moment in time.
//
// ***This function is DEPRECATED. Use NewAmountFromBTC instead.***
func NewAmount(f float64) (Amount, error) {
	return NewAmountFromBTC(f)
}

// NewAmountFromBTC creates an Amount from a floating point value representing
// some value in bitcoin.  NewAmountFromBTC errors if f is NaN or +-Infinity, but
// does not check that the amount is within the total amount of bitcoin
// producible as f may not refer to an amount at a single moment in time.
func NewAmountFromBTC(f float64) (Amount, error) {
	// The amount is only considered invalid if it cannot be represented
	// as an integer type.  This may happen if f is NaN or +-Infinity.
	switch {
	case math.IsNaN(f):
		fallthrough
	case math.IsInf(f, 1):
		fallthrough
	case math.IsInf(f, -1):
		return 0, errors.New("invalid bitcoin amount")
	}

	return round(f * SatoshiPerBitcoin), nil
}
```

wallclockbuilder:
Updated the docs for `btcutil.NewAmount` as recommended and  also squashed everything to make the incremental change cleanly obvious.

amount.go
```diff
@@ -68,6 +68,11 @@ func round(f float64) Amount {
 // some value in bitcoin.  NewAmount errors if f is NaN or +-Infinity, but
 // does not check that the amount is within the total amount of bitcoin
 // producible as f may not refer to an amount at a single moment in time.
+//
+// NewAmount is for specifically for converting BTC to Satoshi.
+// For creating a new Amount with an int64 value which denotes a quantity of Satoshi,
+// do a simple type cast to an Amount.
```
jrick: Nitpick: Go call these type conversions, not casts.

wallclockbuilder: Thanks rick.

davecgh:
OK after rebase/squash.

@jrick: This OK for you now?

jrick:
Yep, looks good.

davecgh:
@wallclockbuilder: Please rebase and squash this and we'll get it merged.  Thanks!

wallclockbuilder:
Squashed.

jrick:
Needs rebase

davecgh:
@wallclockbuilder Thanks.  You squashed it, but still need to rebase it to the latest master.  See the [Rebasing one of your existing pull requests](https://gist.github.com/davecgh/04cb601a147f8c831ca5#rebasing-one-of-your-existing-pull-requests) section of my [btcd Pull Request Management](https://gist.github.com/davecgh/04cb601a147f8c831ca5) gist for details on how to do it if you don't already know.

wallclockbuilder:
Was already working on master so the gist details didn't work for me. Got a noop error on rebase. Fixed it by merging all the recent work from btcsuite into my fork. Its now up to date.

davecgh:
Sorry, but it's still not correct.  We only accept fast-forward merges to master and this PR now has a merge commit in it nor does it merge cleanly.

Probably the easiest way for you to fix it at this point is to reset your master back to the upstream master, cherry-pick the commit back on top, then force push.

> $ git reset --hard 1c7f059
> $ git cherry-pick e55560f
> $ git push -f

This is part of why we suggest working in branches per the gist I linked earlier.

wallclockbuilder:
Thanks. The code snippet is straightforward. 

Noticed you were working on a feature branch. Started doing so myself. Working on a feature branch is less headaches. 
Fixing now.

wallclockbuilder:
Done.
Your feedback was constructive. Thanks.

PR #55, opened 2015-07-28
Return matched tx indices from NewMerkleBlock.

davecgh:
This pull request modifies the `NewMerkleBlock` function to return the matched transaction indices instead of their hashes.  This is being done because it is much easier for the caller to lookup the matched transactions from the original passed block based on their transaction index within the block versus their hashes.


dajohi:
OK

PR #56, opened 2015-08-08
Convert block heights to int32.

davecgh:
This pull request depends on btcsuite/btcd#481.

This pull request converts all block height references to int32 instead of int64.  The current target block production rate is 10 mins per block which means it will take roughly 40,800 years to reach the maximum
height an int32 affords.  Even if the target rate were lowered to one block per minute, it would still take roughly another 4,080 years to reach the maximum.

In the mean time, there is no reason to use a larger type which results in higher memory usage.


dajohi:
OK

PR #58, opened 2015-10-22
Transaction input and output sorting

T909:
Here is the sorting function moved from btcd/wire (previous pull request: btcsuite/btcd#523 )

I have added txsort_test.go but there are no transactions in block 10000 which are "unsorted" (most have only 1 input).  Transaction[1] from block 10001 is unsorted and would be changed by this function, but writing that all in by hand is annoying :) 

Are there other blocks formatted like block 10000 for testing already in btcd?  Or is there some utility to format blocks or txs?  If there is I can put more tests that actually change the txid.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcutil/58)

### DIFF
--- a/txsort.go
+++ b/txsort.go
@@ -1,0 +1,84 @@
+// Copyright (c) 2015 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package btcutil
+
+import (
+	"bytes"
+	"sort"
+
+	"github.com/btcsuite/btcd/wire"
+)
+
+// TxSort
+// Provides functions for sorting tx inputs and outputs according to BIP LI01
+// (https://github.com/kristovatlas/rfc/blob/master/bips/bip-li01.mediawiki)
+
+// TxSort sorts the inputs and outputs of a tx based on BIP LI01
+// It does not modify the transaction given, but returns a new copy
+// which has been sorted and may have a different txid.
+func TxSort(tx *wire.MsgTx) *wire.MsgTx {
+	txCopy := tx.Copy()
+	sort.Sort(sortableInputSlice(txCopy.TxIn))
+	sort.Sort(sortableOutputSlice(txCopy.TxOut))
+	return txCopy
+}
+
+// TxIsSorted checks whether tx has inputs and outputs sorted according
+// to BIP LI01.
+func TxIsSorted(tx *wire.MsgTx) bool {
+	if !sort.IsSorted(sortableInputSlice(tx.TxIn)) {
+		return false
+	}
+	if !sort.IsSorted(sortableOutputSlice(tx.TxOut)) {
+		return false
+	}
+	return true
+}
+
+type sortableInputSlice []*wire.TxIn
+type sortableOutputSlice []*wire.TxOut
+
+// for SortableInputSlice and SortableOutputSlice, three functions are needed
+// to make it sortable with sort.Sort() -- Len, Less, and Swap
+// Len and Swap are trivial.  Less is BIP LI01 specific.
+func (ins sortableInputSlice) Len() int {
+	return len(ins)
+}
+func (outs sortableOutputSlice) Len() int {
+	return len(outs)
+}
+
+func (ins sortableInputSlice) Swap(i, j int) {
+	ins[i], ins[j] = ins[j], ins[i]
+}
+func (outs sortableOutputSlice) Swap(i, j int) {
+	outs[i], outs[j] = outs[j], outs[i]
+}
+
+// Input comparison function.
+// First sort based on input txid (reversed / rpc-style), then index
+func (ins sortableInputSlice) Less(i, j int) bool {
+	ihash := ins[i].PreviousOutPoint.Hash
+	jhash := ins[j].PreviousOutPoint.Hash
+	for b := 0; b < wire.HashSize/2; b++ {
+		ihash[b], ihash[wire.HashSize-1-b] = ihash[wire.HashSize-1-b], ihash[b]
+		jhash[b], jhash[wire.HashSize-1-b] = jhash[wire.HashSize-1-b], jhash[b]
+	}
+	c := bytes.Compare(ihash[:], jhash[:])
+	if c == 0 {
+		// input txids are the same, compare index
+		return ins[i].PreviousOutPoint.Index < ins[j].PreviousOutPoint.Index
+	}
+	return c == -1.
+}
+
+// Output comparison function.
+// First sort based on amount (smallest first), then PkScript
+func (outs sortableOutputSlice) Less(i, j int) bool {
+	if outs[i].Value == outs[j].Value {
+		return bytes.Compare(outs[i].PkScript, outs[j].PkScript) < 0
+	}
+	return outs[i].Value < outs[j].Value
+}
--- a/txsort_test.go
+++ b/txsort_test.go
@@ -1,0 +1,201 @@
+// Copyright (c) 2015 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package btcutil_test
+
+import (
+	"bytes"
+	"encoding/hex"
+	"testing"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+)
+
+// TestSortTx tests SortTx function
+func TestSortTx(t *testing.T) {
+	//	 Use block 100,000 transaction 1.  Already sorted.
+	testTx := Block100000.Transactions[1]
+	sortedTx := btcutil.TxSort(testTx)
+
+	testTxid := testTx.TxSha()
+	sortedTxid := sortedTx.TxSha()
+	if !testTxid.IsEqual(&sortedTxid) {
+		t.Errorf("Sorted TxSha mismatch - got %v, want %v",
+			testTxid.String(), sortedTxid.String())
+	}
+	if !btcutil.TxIsSorted(testTx) {
+		t.Errorf("testTx %v is sorted but reported as unsorted",
+			testTxid.String())
+	}
+
+	// Example 1 0a6a357e2f7796444e02638749d9611c008b253fb55f5dc88b739b230ed0c4c3
+	// test transaction 0a6a357e... which is the first test case of BIPLI01
+	li01Tx1bytes, err := hex.DecodeString(LI01Hex1)
+	if err != nil {
+		t.Errorf("Error in hardcoded hex")
+	}
+
+	var li01Tx1 wire.MsgTx
+	err = li01Tx1.Deserialize(bytes.NewReader(li01Tx1bytes))
+	if err != nil {
+		t.Errorf("Failed to Deserialize li01Tx from byte slice")
+	}
+
+	if btcutil.TxIsSorted(&li01Tx1) {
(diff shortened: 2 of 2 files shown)


davecgh:
Thanks.  As you probably saw, TravisCI is complaining.  Needs a `go fmt` on both files.  See [Code Contribution Checklist](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#61-contribution-checklist).

txsort.go
```diff
@@ -0,0 +1,90 @@
...
+	txCopy := tx.Copy()
+	sortInputs(txCopy.TxIn)
+	sortOutputs(txCopy.TxOut)
+	return txCopy
+}
+
+// TxIsSorted checks whether tx has inputs and outputs sorted according
+// to BIP LI01.
+func TxIsSorted(tx *wire.MsgTx) bool {
+	txCopy := tx.Copy()
```
davecgh: This implementation seem pretty expensive to me.

It would be much more efficient to loop the passed transaction and return false on the first case that violates the sort order.  This would result in a worst case run-time of `O(n)` versus this implementation which is `O(n^2)`.  In addition, it would avoid the copy, avoid the need to actually sort, and avoid the need to take the hash.

In fact, the main purpose for proposing this function was specifically to allow the caller to avoid having to do this.

T909:
Great point there, I modified it.  Using sort.IsSorted() should be faster and the code is cleaner looking too.

txsort.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/btcsuite/btcd/wire"
+)
+
+// TxSort sorts the inputs and outputs of a tx based on BIP LI01
+// It does not modify the transaction given, but returns a new copy
+// which has been sorted and may have a different txid.
+func TxSort(tx *wire.MsgTx) *wire.MsgTx {
+	txCopy := tx.Copy()
+	sortInputs(txCopy.TxIn)
```
davecgh: I'd suggest just doing `sort.Sort(sortableInputSlice(txCopy.TxIn))` here.  In my experience, the Go compiler is not particularly good about inlining.  Also, I noticed the called function is returning a value that isn't used.

txsort.go
```diff
@@ -0,0 +1,100 @@
...
+	sortOutputs(txCopy.TxOut)
```
davecgh: Same here.

txsort.go
```diff
@@ -0,0 +1,100 @@
...
+	return txCopy
+}
+
+// TxIsSorted checks whether tx has inputs and outputs sorted according
+// to BIP LI01.
+func TxIsSorted(tx *wire.MsgTx) bool {
+
+	if !checkSortInputs(tx.TxIn) {
```
davecgh: Same here in regards to just doing `if !sort.IsSorted(sortableInputSlice(tx.TxIn))`.  I don't think there is a readability issue there and it will avoid the function overhead (which likely won't be inlined, unfortunately).

davecgh:
Also, please use the test transactions from the BIP in the tests.  It's important that we adhere to the them.

T909:
OK, that makes sense; I initially wrote it thinking there might be some reason people might want to sort inputs and outputs separately but I can't actually think of a reason for that.  I've condensed it here.

Also for the "block 10000" var in block_test.go, how did you guys create that?  Just by hand?  I guess I could use that as a template for the BIP test cases, but an automated tool to make that would be a lot faster.

davecgh:
A relatively simple way to test against the transactions in the BIP is to pull the raw hex for the transactions, then make use of `hex.DecodeString` and the following to turn it into a transaction:

```Go
	serializedTx, err := hex.DecodeString("....")
	if err != nil {
		t.Errorf(...)
	}

	var tx wire.MsgTx
	err = tx.Deserialize(bytes.NewReader(serializedTx))
	if err != nil {
		t.Errorf(...)
	}

	sortedTx := btcutil.TxSort(&tx)

	// Compare against the expected inputs and outputs...
```

T909:
OK I will put some giant hex lines in and use hex.DecodeString() :)

davecgh:
You can use http://webbtc.com/tx/0a6a357e2f7796444e02638749d9611c008b253fb55f5dc88b739b230ed0c4c3.hex to pull the transaction hex.

davecgh:
If you'd prefer not to hard code the hex in the code, you could load it from a file too and put the file in a `testdata` directory.  See https://github.com/btcsuite/btcd/tree/master/blockchain/testdata for example.

davecgh:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.
Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcutil/58)

davecgh:
Reviewed 1 of 2 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[txsort_test.go, line 30 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcutil/58#-K1H0UBWmeD_QgaCJeKg)** ([raw file](https://github.com/btcsuite/btcutil/blob/f2d621d49a83df92532ab4b6ba1122f45875146d/txsort_test.go#L30)):</sup>
Standard Go style is to use lowercase for unexported variables.  So this should be `li01Tx1Bytes`.  The same is true through the rest of this file.


---

<sup>**[txsort_test.go, line 60 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcutil/58#-K1H1HUiy-hHy3PfAknI)** ([raw file](https://github.com/btcsuite/btcutil/blob/f2d621d49a83df92532ab4b6ba1122f45875146d/txsort_test.go#L60)):</sup>
I'd suggest adding the following helper that I pulled from the `chaincfg` tests:

```Go
// newShaHashFromStr converts the passed big-endian hex string into a
// wire.ShaHash.  It only differs from the one available in wire in that
// it panics on an error since it will only (and must only) be called with
// hard-coded, and therefore known good, hashes.
func newShaHashFromStr(hexStr string) *wire.ShaHash {
	sha, err := wire.NewShaHashFromStr(hexStr)
	if err != nil {
		panic(err)
	}
	return sha
}
```

Then you can just do `wantSha := newShaHashFromStr("0a6a...")` here.   Same goes for below.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcutil/58)

davecgh:
Also, as can been seen from [Coveralls](https://coveralls.io/builds/3925776/source?filename=txsort.go), there is no coverage for the case where the outputs need to be reordered.

davecgh:
Speaking of which, we probably want to let the author of the BIP know there isn't a reference transaction which causes a reorder in the outputs.  It seems me me like it should have one of those...

davecgh:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[txsort.go, line 1 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcutil/58#-K1H8znSR_DCunYrhoeB)** ([raw file](https://github.com/btcsuite/btcutil/blob/f2d621d49a83df92532ab4b6ba1122f45875146d/txsort.go#L1)):</sup>
Please add the standard license header with a copyright year of 2015.


---

<sup>**[txsort_test.go, line 1 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcutil/58#-K1H96y1Pimjfiopcnuz)** ([raw file](https://github.com/btcsuite/btcutil/blob/f2d621d49a83df92532ab4b6ba1122f45875146d/txsort_test.go#L1)):</sup>
Please add the standard license header with a copyright year of 2015.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcutil/58)

davecgh:
Reviewed 1 of 2 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcutil/58)

davecgh:
Reviewed 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[txsort.go, line 1 \[r6\]](https://reviewable.io:443/reviews/btcsuite/btcutil/58#-K1IIFA2vjROvyu-Tf1j)** ([raw file](https://github.com/btcsuite/btcutil/blob/d747b7d7c42fa0519df46d5120ddc4b2d55c65dc/txsort.go#L1)):</sup>
2015.  It's all new!


---

<sup>**[txsort_test.go, line 1 \[r6\]](https://reviewable.io:443/reviews/btcsuite/btcutil/58#-K1IISes4Acf-VXt3LvN)** ([raw file](https://github.com/btcsuite/btcutil/blob/d747b7d7c42fa0519df46d5120ddc4b2d55c65dc/txsort_test.go#L1)):</sup>
Same as other file.  2015.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcutil/58)

davecgh:
I believe that covers it.  Looks good to me.  I might do a separate PR to move the hex out to files in a `testdata` directory just because I think that would be cleaner, but that's no reason to hold this PR up.

OK


---

Reviewed 2 of 2 files at r7.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcutil/58)

davecgh:
Oh, please do rebase and squash everything into a single commit.  I forgot to mention that.

davecgh:
[Model Git Commit Message](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#44-code-documentation-and-commenting)

davecgh:
Thanks!  Looks good.

T909:
Thanks a lot, sorry it was a bit of a mess!  I think I've got the idea now though :)

davecgh:
No worries.  We try to keep the code quality high, so I know it can a bit of a learning curve.  The end result is worth it though.

davecgh:
Also, I won't hold this PR up for it, but for future reference, a neat way to do tests is to test-driven use anon structs.  You could've done like:

```Go
	tests := []struct {
		name     string
		txHex    string
		isSorted bool
		wantSha  string
	}{
		{
			name:     "first tx from BIP",
			txHex:    li01Tx1Hex,
			isSorted: false,
			wantSha:  newShaHashFromStr("...."),
		},
		{
			name:     "second tx from BIP",
			txHex:    li02Tx1Hex,
			isSorted: false,
			wantSha:  newShaHashFromStr("...."),
		},
		// etc...
	}
	for _, test := range tests {
		var tx wire.MsgTx
		err := tx.Deserialize(bytes.NewReader(test.txHex))
		if err != nil {
			t.Errorf("Deserialize (%s): unexpected error %v",
				test.name, err)
			continue
		}

		if got := btcutil.TxIsSorted(&tx); got != test.isSorted {
			t.Errorf("TxIsSorted (%s): sort does not matched "+
				"expected - got %v, want %v", test.name, got,
				test.isSorted)
			continue
		}

		// And so on...
	}
```

davecgh:
For reference, I'm going to merge this, but I'll be pushing a couple of PRs to make it a separate package and separate and convert the tests as discussed.

PR #60, opened 2015-10-23
txsort: Convert tests, optimize, and cleanup code.

davecgh:
- Move hex for test txns into separate files in the testdata directory
- Convert tests to table-driven tests
- Make comments more consistent with the rest of the codebase
- Optimize the input sorting function to perform the hash equivalence check before reversing the bytes so it can be avoided in that case

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcutil/60)


davecgh:
Coveralls failure is a false positive.  The overall coverage of the entire btcutil repo went down as a result of the number of lines changed, however the `txsort` still has 100% coverage.

jrick:
ok

PR #61, opened 2015-10-23
txsort: Add InPlaceSort function.

davecgh:
This adds a new function with loud warnings which allows sorting a transaction in place by mutating it.  This is more efficient for the caller if they are the ones creating the transaction and are sure it will not invalid any cache or containing structures.

The Sort function which makes a copy and is therefore does not mutate the passed transaction is still available and the default method.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcutil/61)


davecgh:
Coveralls failure is a false positive.  The overall coverage of the entire btcutil repo went down as a result of the number of lines changed, however the `txsort` package still has 100% coverage.

jrick:
ok

PR #67, opened 2016-02-09
hdkeychain: Correct extended privkey serialization.

davecgh:
This corrects an issue with the serialization of extended private keys where certain underlying derivations could cause lead to printing extended privkeys that did not have the expected xprv prefix.

In addition, tests for private key derivation have been added as well as a specific test which triggers the previously failing case.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcutil/67)


jrick:
ok

PR #68, opened 2016-02-26
Fix range check in bloom filter false positive rate

T909:
Tiny fix here.  If you gave an argument of 0 for the false positive rate in NewFilter(), it would make a 0-length filter, always matching.  That's really a false positive rate of 1 which seems like the opposite effect.  Interpreting a 0 as the lower limit of 1e-9 seems more intuitive behavior when receiving an out-of-bounds argument.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcutil/68)

### DIFF
--- a/bloom/filter_test.go
+++ b/bloom/filter_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2013, 2014 The btcsuite developers
+// Copyright (c) 2013, 2014, 2016 The btcsuite developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
@@ -92,6 +92,66 @@ func TestFilterInsert(t *testing.T) {
 		t.Errorf("TestFilterInsert failure: got %v want %v\n",
 			got.Bytes(), want)
 		return
+	}
+}
+
+// TestFilterFPRange checks that new filters made with out of range
+// false positive targets result in either max or min false positive rates.
+func TestFilterFPRange(t *testing.T) {
+	tests := []struct {
+		name   string
+		hash   string
+		want   string
+		filter *bloom.Filter
+	}{
+		{
+			name:   "fprates > 1 should be clipped at 1",
+			hash:   "02981fa052f0481dbc5868f4fc2166035a10f27a03cfd2de67326471df5bc041",
+			want:   "00000000000000000001",
+			filter: bloom.NewFilter(1, 0, 20.9999999769, wire.BloomUpdateAll),
+		},
+		{
+			name:   "fprates less than 1e-9 should be clipped at min",
+			hash:   "02981fa052f0481dbc5868f4fc2166035a10f27a03cfd2de67326471df5bc041",
+			want:   "0566d97a91a91b0000000000000001",
+			filter: bloom.NewFilter(1, 0, 0, wire.BloomUpdateAll),
+		},
+		{
+			name:   "negative fprates should be clipped at min",
+			hash:   "02981fa052f0481dbc5868f4fc2166035a10f27a03cfd2de67326471df5bc041",
+			want:   "0566d97a91a91b0000000000000001",
+			filter: bloom.NewFilter(1, 0, -1, wire.BloomUpdateAll),
+		},
+	}
+
+	for _, test := range tests {
+		// Convert test input to appropriate types.
+		hash, err := wire.NewShaHashFromStr(test.hash)
+		if err != nil {
+			t.Errorf("NewShaHashFromStr unexpected error: %v", err)
+			continue
+		}
+		want, err := hex.DecodeString(test.want)
+		if err != nil {
+			t.Errorf("DecodeString unexpected error: %v\n", err)
+			continue
+		}
+
+		// Add the test hash to the bloom filter and ensure the
+		// filter serializes to the expected bytes.
+		f := test.filter
+		f.AddShaHash(hash)
+		got := bytes.NewBuffer(nil)
+		err = f.MsgFilterLoad().BtcEncode(got, wire.ProtocolVersion)
+		if err != nil {
+			t.Errorf("BtcDecode unexpected error: %v\n", err)
+			continue
+		}
+		if !bytes.Equal(got.Bytes(), want) {
+			t.Errorf("serialized filter mismatch: got %x want %x\n",
+				got.Bytes(), want)
+			continue
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


davecgh:
Nice catch.  Would you please also add a test that fails the current code and passes the new code to be sure any future regressions in this area are prevented?  I believe there is already a test for the maximum size filter, but there is obviously not one for the minimum case.

Also, please update the copyright year on the touched files.

Thanks!

T909:
Sure, I added a short test case and added 2016 to the copyright years at top.  Let me know if there's anything else.

bloom/filter_test.go
```diff
@@ -95,6 +95,68 @@ func TestFilterInsert(t *testing.T) {
...
+func TestFilterFPRange(t *testing.T) {
+
+	// fprates greater than 1 should be clipped at 1, resulting in a high
+	// false positive rate and small filter size.
+	f := bloom.NewFilter(1, 0, 20.9999999769, wire.BloomUpdateAll)
+
+	inputStr := "02981fa052f0481dbc5868f4fc2166035a10f27a03cfd2de67326471df5bc041"
+	sha, err := wire.NewShaHashFromStr(inputStr)
+	if err != nil {
+		t.Errorf("TestMerkleBlockP2PubKeyOnly NewShaHashFromStr failed: %v", err)
```
davecgh: The function name in the error is incorrect, but it isn't really needed in top level test functions like this since the `go test` testing framework will print the function for you already.

Same goes for the others.

bloom/filter_test.go
```diff
@@ -95,6 +95,68 @@ func TestFilterInsert(t *testing.T) {
...
+	}
+	f.AddShaHash(sha)
+
+	want, err := hex.DecodeString("00000000000000000001")
+	if err != nil {
+		t.Errorf("TestFilterFPRange DecodeString failed: %v\n", err)
+		return
+	}
+	got := bytes.NewBuffer(nil)
+	err = f.MsgFilterLoad().BtcEncode(got, wire.ProtocolVersion)
```
davecgh: This is all a repeat of the same code.  I'd rather this be refactored to use a table driven test as it's easier to add more cases that way as well.

bloom/filter_test.go
```diff
@@ -95,6 +95,68 @@ func TestFilterInsert(t *testing.T) {
...
+	}
+
+	// fprates less than 1e-9, including 0 and negative numbers, should clip at the
+	// minimum false positive rate of 1e-9, resulting in a larger filter.
+
+	f = bloom.NewFilter(1, 0, 0, wire.BloomUpdateAll)
+	f.AddShaHash(sha)
+
+	want, err = hex.DecodeString(
+		"000000000000000000010566d97a91a91b0000000000000001")
```
davecgh: The buffer isn't being reset, so this is checking against the concatenated bytes of both filters instead of only the filter under test.

bloom/filter_test.go
```diff
@@ -95,6 +95,68 @@ func TestFilterInsert(t *testing.T) {
...
+	}
+
+	err = f.MsgFilterLoad().BtcEncode(got, wire.ProtocolVersion)
+	if err != nil {
+		t.Errorf("TestFilterFPRange BtcDecode failed: %v\n", err)
+		return
+	}
+
+	if !bytes.Equal(got.Bytes(), want) {
+		t.Errorf("TestFilterFPRange failure: got %v want %v\n",
```
davecgh: Bytes slices should be printed with `%x` instead of `%v` since hex is much more readable.

bloom/filter_test.go
```diff
@@ -95,6 +95,68 @@ func TestFilterInsert(t *testing.T) {
 	}
 }
 
+// TestFilterFPRange checks that new filters made with out of range
+// false positive targets result in either max or min false positive rates.
+func TestFilterFPRange(t *testing.T) {
```
davecgh: Actually, I'll go ahead and refactor this along with the mentioned thing and paste the refactored version here in as it'll only take a few minutes.

davecgh: As mentioned, here is the refactored version.  I also went ahead and added a test with a negative fprate.

```Go
func TestFilterFPRange(t *testing.T) {
	tests := []struct {
		name   string
		hash   string
		want   string
		filter *bloom.Filter
	}{
		{
			name:   "fprates > 1 should be clipped at 1",
			hash:   "02981fa052f0481dbc5868f4fc2166035a10f27a03cfd2de67326471df5bc041",
			want:   "00000000000000000001",
			filter: bloom.NewFilter(1, 0, 20.9999999769, wire.BloomUpdateAll),
		},
		{
			name:   "fprates less than 1e-9 should be clipped at min",
			hash:   "02981fa052f0481dbc5868f4fc2166035a10f27a03cfd2de67326471df5bc041",
			want:   "0566d97a91a91b0000000000000001",
			filter: bloom.NewFilter(1, 0, 0, wire.BloomUpdateAll),
		},
		{
			name:   "negative fprates should be clipped at min",
			hash:   "02981fa052f0481dbc5868f4fc2166035a10f27a03cfd2de67326471df5bc041",
			want:   "0566d97a91a91b0000000000000001",
			filter: bloom.NewFilter(1, 0, -1, wire.BloomUpdateAll),
		},
	}

	for _, test := range tests {
		// Convert test input to appropriate types.
		hash, err := wire.NewShaHashFromStr(test.hash)
		if err != nil {
			t.Errorf("NewShaHashFromStr unexpected error: %v", err)
			continue
		}
		want, err := hex.DecodeString(test.want)
		if err != nil {
			t.Errorf("DecodeString unexpected error: %v\n", err)
			continue
		}

		// Add the test hash to the bloom filter and ensure the
		// filter serializes to the expected bytes.
		f := test.filter
		f.AddShaHash(hash)
		got := bytes.NewBuffer(nil)
		err = f.MsgFilterLoad().BtcEncode(got, wire.ProtocolVersion)
		if err != nil {
			t.Errorf("BtcDecode unexpected error: %v\n", err)
			continue
		}
		if !bytes.Equal(got.Bytes(), want) {
			t.Errorf("serialized filter mismatch: got %x want %x\n",
				got.Bytes(), want)
			continue
		}
	}
}
```

T909:
OK thanks, yeah I had just copied from other tests in the file without looking into them too much.  I put your code as a commit in my fork, is that the best way to do it?

davecgh:
Yep, that's fine.  We always squash PRs down to a single commit anyways before merging so the commit history doesn't end up cluttered with all the working revisions during PR review.  Speaking of which, this looks good to go now.  Please rebase/squash it to a single commit and I'll get it merged.

Thanks for your work on this.  I always appreciate your fixes and contributions.

T909:
OK thanks! Squahed into 1 commit.
Also working with @Roasbeef on a segregated witness PR.  Not ready yet and that's a pretty big one :)

PR #72, opened 2016-03-22
Update TravisCI for Go 1.6.

davecgh:
Now that Go 1.6 has been released, remove the old Go versions from the configurations tested by TravisCI and add the latest point releases.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcutil/72)


dajohi:
OK

PR #73, opened 2016-03-22
bloom: Correct merkle block test error print.

davecgh:
The test would have erroneously printed the function address instead of the received bytes if it failed.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcutil/73)


dajohi:
OK

davecgh:
For reference, the Travis failure is due to it testing an outdated version of Go and is not related to this PR.

The Travis issue is resolved by #72.

PR #75, opened 2016-04-10
Add utility functions for blocks and transactions witness data 

Roasbeef:
This PR adds a number of utility messages that will be useful for callers interacting with blocks and transactions that contain witness data: 
- The first new method `btcutil.Block.BytesNoWitness` provides a helper message to generate a serialized version of a block _without_ any witness data (if any). 
- The second new method `btcutil.Tx.WitnessHash` is equivelant to `tx.Hash` but instead returns the `wtxid` of a transaction. 
- The finally added method is `btcutil.Tx.HasWitness`, which is ideneitcal to `tx.HasWitness`, but caches the results. This is a useful performance optimization as a transaction may potentially have thousands of inputs, so we avoid interacting over them repeatedly during block validation and other contexts. 

NOTE: This PR requires btcsuite/btcd#656

### DIFF
--- a/block.go
+++ b/block.go
@@ -32,12 +32,13 @@ func (e OutOfRangeError) Error() string {
 // transactions on their first access so subsequent accesses don't have to
 // repeat the relatively expensive hashing operations.
 type Block struct {
-	msgBlock        *wire.MsgBlock  // Underlying MsgBlock
-	serializedBlock []byte          // Serialized bytes for the block
-	blockHash       *chainhash.Hash // Cached block hash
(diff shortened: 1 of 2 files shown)


halseth:
Bech32 PR is up #100

Roasbeef:
Modified this PR to eliminate the segwit address portion as that portion will now be implemented by #100.

davecgh:
Would you update the title and descriptions please?

davecgh:
Need a rebase and then I'll get this merged.

PR #78, opened 2016-08-07
Update for recent chainhash-related API changes.

davecgh:
**This PR requires btcsuite/btcd#729.**

This updates all code in the main package and subpackages to make use of the new `chainhash` package since the old `wire.ShaHash` type and functions have been removed in favor of the abstracted package.

Also, since this required API changes anyways and the hash algorithm is no longer tied specifically to SHA, all other functions throughout the code base which had `Sha` in their name have been changed to `Hash` so they are not incorrectly implying the hash algorithm.

The following is an overview of the changes:

- Update all references to `wire.ShaHash` to the new `chainhash.Hash` type
- Rename the following functions and update all references:
  - `Block.Sha` -> `Hash`
  - `Block.TxSha` -> `TxHash`
  - `Tx.Sha` -> `Hash`
  - `bloom.Filter.AddShaHash` -> `AddHash`
- Rename all variables that included `sha` in their name to include `hash` instead
- Add license headers to `coinset` package files

**NOTE:** This PR will fail the CI build because it depends on the aforementioned btcd PR.

### DIFF
--- a/block.go
+++ b/block.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2013-2014 The btcsuite developers
+// Copyright (c) 2013-2016 The btcsuite developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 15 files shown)


jcvernaleo:
OK

marcopeereboom:
ok

cjepson:
OK

PR #83, opened 2016-11-03
Modify goclean to use gometalint like btcd

jcvernaleo:


goclean.sh
```diff
@@ -5,14 +5,19 @@
 # 3. golint        (https://github.com/golang/lint)
 # 4. go vet        (http://golang.org/cmd/vet)
 # 5. test coverage (http://blog.golang.org/cover)
+#
+# gometaling (github.com/alecthomas/gometalinter) is used to run each each
```
davecgh: gometalint

jcvernaleo: Fixed that typo.

goclean.sh
```diff
@@ -5,14 +5,19 @@
...
-test -z "$(gofmt -l -w .     | tee /dev/stderr)"
-test -z "$(goimports -l -w . | tee /dev/stderr)"
-test -z "$(golint .          | tee /dev/stderr)"
-go vet ./...
+test -z "$(gometalinter --disable-all \
+--enable=gofmt \
+--enable=golint \
+--enable=vet \
+--enable=goimports \
+--deadline=120s ./... | grep -v 'ALL_CAPS\|OP_\|ExampleNew' 2>&1 | tee /dev/stderr)"
```
dajohi: The grep should be removed.

jcvernaleo: I removed the ALL_CAPS and OP_ though since those were specific to part of btcd.

dajohi:
OK

goclean.sh
```diff
@@ -5,14 +5,19 @@
...
 # Automatic checks
-test -z "$(gofmt -l -w .     | tee /dev/stderr)"
-test -z "$(goimports -l -w . | tee /dev/stderr)"
-test -z "$(golint .          | tee /dev/stderr)"
-go vet ./...
+test -z "$(gometalinter --disable-all \
+--enable=gofmt \
+--enable=golint \
+--enable=vet \
+--enable=goimports \
```
davecgh: Minor nitpick:

This should be in the same order as the comments above.

davecgh: More specifically:

```
--enable=gofmt \
--enable=goimports \
--enable=golint \
--enable=vet \
```

jcvernaleo: Fixed (and I agree, the order should match).

PR #87, opened 2016-11-23
build: Add unconvert linter to goclean.sh

dajohi:


jcvernaleo:
#83 already does this (and passes travis) so there is no reason for this PR.

dajohi:
#83 is missing a unconvert and gosimple.

jcvernaleo:
Those should be added as a seperate PR and not part of the first one so I still do not think this one should go in.  That is the way we did it for the other repo

davecgh:
I will be merging #83.  Would you please split the addition of `unconvert` and `gosimple` into separate PRs on top of #83 much like it was done in btcd?

davecgh:
Alright #83 has been merged.  Please rebase this one and redact it such that it only contains the `unconvert` changes and update the commit message and PR title accordingly.  Then a separate PR for `gosimple`, so it's consistent with the other repos.

dajohi:
Updated.

PR #91, opened 2017-01-10
Drop fastsha256 in favor of crypto/sha256

dajohi:


stevenroose:
Was fastsha256 no longer faster?

dajohi:
if my memory serves, the fastsha256 pkg was created because we both had the assembly before it was in a release and needed to provide midstate for the getwork RPC.

crypto/sha256 has had many optimizations and keeping fastsha256 in sync is tedious just to support the deprecated getwork RPC.

By dropping support for the getwork RPC, we no longer need fastsha256.  By switching to crypto/sha256, we also get the better performing code.

See https://github.com/btcsuite/btcd/pull/894

davecgh:
It was originally only forked from the stdlib version for a couple of reasons.

1) The btcsuite project originally paid to have the work performed on optimizing the Go stdlib package to increase its performance since the existing code was really slow, however, since btcsuite supports both the most recent and the previous version of Go, it was necessary to create and use a separate package in the interim which led to the birth of fastsha256
2)  The `getwork` RPC requires access to the sha256 midstate of the first round which the stdlib version does not provide.

Now that `getwork` is being removed and several new versions of Go have been released, there is no longer any need for the separate package.

As an additional benefit, the stdlib version is actually a little bit faster now too since it has had some additional work done on it and we hadn't synced it up in a while.

PR #92, opened 2017-02-16
Switch to upstream golang.org/x/crypto

dajohi:


jrick:
We need to do this in all of the repos to avoid pulling in multiple versions of the same package under different names.

dajohi:
will do.

PR #95, opened 2017-05-05
hdkeychain: BIP32 maximum depth is 255

afk11:
BIP32 keys serialize the depth as a uint8 over the wire.  I noticed
uint16 was being used and that the depth was being taken modulo 256
during serialization.

This seems like a bug, as the behaviour is not described in the BIP,
and also introduces incompatibilities which can be hard to make sense
of.  For example, the parent fingerprint should be 0x00000000 for a key
of depth zero, whereas with the existing code if depth=256, then the
serialization will specify 0 depth but still set a parent fingerprint.

### DIFF
--- a/hdkeychain/extendedkey.go
+++ b/hdkeychain/extendedkey.go
@@ -49,6 +49,9 @@ const (
 	// fingerprint, 4 bytes child number, 32 bytes chain code, and 33 bytes
 	// public/private key data.
 	serializedKeyLen = 4 + 1 + 4 + 4 + 32 + 33 // 78 bytes
+
+	// maxUint8 is the max positive integer which can be serialized in a uint8
+	maxUint8 = 1<<8 - 1
 )
 
 var (
@@ -56,6 +59,11 @@ var (
 	// attempted to derive a hardened extended key from a public key.
 	ErrDeriveHardFromPublic = errors.New("cannot derive a hardened key " +
 		"from a public key")
+
+	// ErrDeriveBeyondMaxDepth describes an error in which the caller
+	// has attempted to derive more than 255 keys from a root key.
+	ErrDeriveBeyondMaxDepth = errors.New("cannot derive a key with more than " +
+		"255 indices in its path")
 
 	// ErrNotPrivExtKey describes an error in which the caller attempted
 	// to extract a private key from a public extended key.
@@ -101,7 +109,7 @@ type ExtendedKey struct {
 	key       []byte // This will be the pubkey for extended pub keys
 	pubKey    []byte // This will only be set for extended priv keys
 	chainCode []byte
-	depth     uint16
+	depth     uint8
 	parentFP  []byte
 	childNum  uint32
 	version   []byte
@@ -111,7 +119,7 @@ type ExtendedKey struct {
 // newExtendedKey returns a new instance of an extended key with the given
 // fields.  No error checking is performed here as it's only intended to be a
 // convenience method used to create a populated struct.
-func newExtendedKey(version, key, chainCode, parentFP []byte, depth uint16,
+func newExtendedKey(version, key, chainCode, parentFP []byte, depth uint8,
 	childNum uint32, isPrivate bool) *ExtendedKey {
 
 	// NOTE: The pubKey field is intentionally left nil so it is only
@@ -194,6 +202,10 @@ func (k *ExtendedKey) Child(i uint32) (*ExtendedKey, error) {
 	// 2) Private extended key -> Non-hardened child private extended key
 	// 3) Public extended key -> Non-hardened child public extended key
 	// 4) Public extended key -> Hardened child public extended key (INVALID!)
+
+	if k.depth == maxUint8 {
+		return nil, ErrDeriveBeyondMaxDepth
+	}
 
 	// Case #4 is invalid, so error out early.
 	// A hardened child extended key may not be created from a public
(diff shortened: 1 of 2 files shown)


afk11:
Apparently coverage went _down_.. Unsure precisely why, but let me know if it's an issue.

hdkeychain/extendedkey.go
```diff
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"math/big"
 
+	"math"
```
stevenroose: Remove the empty line so that goimports sorts them correctly.

hdkeychain/extendedkey_test.go
```diff
@@ -15,6 +15,8 @@ import (
 	"reflect"
 	"testing"
 
+	"math"
```
stevenroose: Ditto.

hdkeychain/extendedkey_test.go
```diff
@@ -1013,3 +1014,38 @@ func TestZero(t *testing.T) {
 		}
 	}
 }
+
+func TestExtendedKey_MaximumDepth(t *testing.T) {
```
davecgh: Please just call this `TestMaximumDepth`.  Using snake case is discouraged by Effective Go guidelines.

Also, it's missing a comment to be consistent with the rest of the test functions.  The following should do if you don't already have something in mind:

```Go
// TestMaximumDepth ensures that attempting to retrieve a child key when already
// at the maximum depth is not allowed.
```

hdkeychain/extendedkey.go
```diff
@@ -195,6 +201,10 @@ func (k *ExtendedKey) Child(i uint32) (*ExtendedKey, error) {
 	// 3) Public extended key -> Non-hardened child public extended key
 	// 4) Public extended key -> Hardened child public extended key (INVALID!)
 
+	if k.depth == math.MaxUint8 {
```
davecgh: Rather than importing the entire `math` package for this sole constant, I would just prefer an unexported constant in the constants block near the top of the file.

```Go
maxUint8 = 1<<8 - 1
```

Then this would become:

```Go
if k.depth == maxUint8 {
```

The tests can still pull in the math package since they don't affect the primary package.

hdkeychain/extendedkey_test.go
```diff
@@ -1013,3 +1014,38 @@ func TestZero(t *testing.T) {
...
+
+	extKey, err := hdkeychain.NewMaster([]byte(`abcd1234abcd1234abcd1234abcd1234`), &chaincfg.MainNetParams)
+	if err != nil {
+		t.Error("MaxDepthTest: Failed to produce test fixture key from string")
+		return
+	}
+
+	var i uint8
+	for i = 0; i < math.MaxUint8; i++ {
+		newKey, err := extKey.Child(uint32(1))
```
davecgh: No need for uint32 cast of an integer constant.

hdkeychain/extendedkey_test.go
```diff
@@ -1013,3 +1014,38 @@ func TestZero(t *testing.T) {
...
+		return
+	}
+
+	if _, ok := err.(error); !ok {
+		t.Error("MaxDepthTest: Deriving 256th key should not succeed")
+		return
+	}
+
+	if err != hdkeychain.ErrDeriveBeyondMaxDepth {
+		t.Error("MaxDepthTest: Received unexpected error during est")
```
davecgh: est -> test

hdkeychain/extendedkey_test.go
```diff
@@ -1013,3 +1014,38 @@ func TestZero(t *testing.T) {
...
+func TestExtendedKey_MaximumDepth(t *testing.T) {
+
+	extKey, err := hdkeychain.NewMaster([]byte(`abcd1234abcd1234abcd1234abcd1234`), &chaincfg.MainNetParams)
+	if err != nil {
+		t.Error("MaxDepthTest: Failed to produce test fixture key from string")
+		return
+	}
+
+	var i uint8
+	for i = 0; i < math.MaxUint8; i++ {
```
davecgh: `i` is not needed outside of the loop.

```Go
for i := uint8(0); i < math.MaxUint8; i++ {
```

hdkeychain/extendedkey_test.go
```diff
@@ -1013,3 +1014,38 @@ func TestZero(t *testing.T) {
...
+		extKey = newKey
+	}
+
+	noKey, err := extKey.Child(uint32(1))
+	if noKey != nil {
+		t.Error("MaxDepthTest: Deriving 256th key should not succeed")
+		return
+	}
+
+	if _, ok := err.(error); !ok {
```
davecgh: This is not needed.  The check for `err != hdkeychain.ErrDeriveBeyondMaxDepth` below is sufficient.

hdkeychain/extendedkey_test.go
jrick: The standard library is littered with tests with _ in the names.  It really doesn't matter.

davecgh: Well none of the rest of the code base uses them, except for examples where it is required to make them show up properly on godoc, so I'd like to keep the consistency regardless.

jrick:
```
PS C:\Go\src> gci -r *_test.go | sls -ca 'func Test.*_.*\(' | measure


Count    : 191
Average  :
Sum      :
Maximum  :
Minimum  :
Property :
```

:)

davecgh:
```
$ pwd
/go/src/github.com/btcsuite/btcutil
$ grep 'func Test.*_.*' $(find . -name '*_test.go' | xargs)
$
```

afk11:
Hey all, thanks for the feedback, pushed some commits now to address these, will squash once tests pass.

PR #96, opened 2017-05-09
hdkeychain: Consolidate tests into package.

davecgh:
Putting the test code in the same package makes it easier for forks since they don't have to change the import paths as much.

Also, address a few style and consistent nits while here:
- Prefer `t.Fatalf` over `t.Errorf` followed by a return
- Use the consistent style of starting a test function comments with the test name
- Prefix test errors by the function being called instead of the one doing the calling since the caller itself is already logged by the test framework
- Check err in max depth test before checking the returned key is nil

### DIFF
--- a/hdkeychain/extendedkey_test.go
+++ b/hdkeychain/extendedkey_test.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
-package hdkeychain_test
+package hdkeychain
 
 // References:
 //   [BIP32]: BIP0032 - Hierarchical Deterministic Wallets
@@ -17,7 +17,6 @@ import (
 	"testing"
 
 	"github.com/btcsuite/btcd/chaincfg"
-	"github.com/btcsuite/btcutil/hdkeychain"
 )
 
 // TestBIP0032Vectors tests the vectors provided by [BIP32] to ensure the
@@ -196,7 +195,7 @@ tests:
 			continue
 		}
 
-		extKey, err := hdkeychain.NewMaster(masterSeed, test.net)
+		extKey, err := NewMaster(masterSeed, test.net)
 		if err != nil {
 			t.Errorf("NewMaster #%d (%s): unexpected error when "+
 				"creating new master key: %v", i, test.name,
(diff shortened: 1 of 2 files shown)


hdkeychain/extendedkey_test.go
```diff
@@ -778,41 +777,38 @@ func TestNet(t *testing.T) {
 func TestErrors(t *testing.T) {
 	// Should get an error when seed has too few bytes.
 	net := &chaincfg.MainNetParams
-	_, err := hdkeychain.NewMaster(bytes.Repeat([]byte{0x00}, 15), net)
-	if err != hdkeychain.ErrInvalidSeedLen {
-		t.Errorf("NewMaster: mismatched error -- got: %v, want: %v",
-			err, hdkeychain.ErrInvalidSeedLen)
+	_, err := NewMaster(bytes.Repeat([]byte{0x00}, 15), net)
+	if err != ErrInvalidSeedLen {
+		t.Fatalf("NewMaster: mismatched error -- got: %v, want: %v",
```
jrick: there was no removed error return here, why the switch to Fatalf?

hdkeychain/extendedkey_test.go
```diff
@@ -778,41 +777,38 @@ func TestNet(t *testing.T) {
...
 	}
 
 	// Should get an error when seed has too many bytes.
-	_, err = hdkeychain.NewMaster(bytes.Repeat([]byte{0x00}, 65), net)
-	if err != hdkeychain.ErrInvalidSeedLen {
-		t.Errorf("NewMaster: mismatched error -- got: %v, want: %v",
-			err, hdkeychain.ErrInvalidSeedLen)
+	_, err = NewMaster(bytes.Repeat([]byte{0x00}, 65), net)
+	if err != ErrInvalidSeedLen {
+		t.Fatalf("NewMaster: mismatched error -- got: %v, want: %v",
```
jrick: same

hdkeychain/extendedkey_test.go
davecgh: There should have been a return there as this was intended to be Fatal.

hdkeychain/extendedkey_test.go
davecgh: Same.

PR #100, opened 2017-06-02
Add Bech32 address format package.

halseth:
This PR implements the Bech32 address format as specified in BIP 173: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki

Test are modeled after the Python reference implementation (https://github.com/sipa/bech32/blob/master/ref/python/tests.py) to ensure compatibility.

### DIFF
--- a/address.go
+++ b/address.go
@@ -5,14 +5,34 @@
 package btcutil
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
+	"fmt"
+	"strings"
 
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcutil/base58"
+	"github.com/btcsuite/btcutil/bech32"
 	"golang.org/x/crypto/ripemd160"
 )
+
+// UnsupportedWitnessVerError describes an error where a segwit address being
+// decoded has an unsupported witness version.
+type UnsupportedWitnessVerError byte
+
+func (e UnsupportedWitnessVerError) Error() string {
+	return "unsupported witness version: " + string(e)
+}
+
+// UnsupportedWitnessProgLenError describes an error where a segwit address
+// being decoded has an unsupported witness program length.
+type UnsupportedWitnessProgLenError int
+
+func (e UnsupportedWitnessProgLenError) Error() string {
+	return "unsupported witness program length: " + string(e)
+}
 
 var (
 	// ErrChecksumMismatch describes an error where decoding failed due
@@ -42,6 +62,39 @@ func encodeAddress(hash160 []byte, netID byte) string {
 	// Format is 1 byte for a network and address class (i.e. P2PKH vs
 	// P2SH), 20 bytes for a RIPEMD160 hash, and 4 bytes of checksum.
 	return base58.CheckEncode(hash160[:ripemd160.Size], netID)
+}
+
+// encodeSegWitAddress creates a bech32 encoded address string representation
+// from witness version and witness program.
+func encodeSegWitAddress(hrp string, witnessVersion byte, witnessProgram []byte) (string, error) {
+	// Group the address bytes into 5 bit groups, as this is what is used to
+	// encode each character in the address string.
+	converted, err := bech32.ConvertBits(witnessProgram, 8, 5, true)
+	if err != nil {
+		return "", err
+	}
+
+	// Concatenate the witness version and program, and encode the resulting
+	// bytes using bech32 encoding.
+	combined := make([]byte, len(converted)+1)
+	combined[0] = witnessVersion
+	copy(combined[1:], converted)
+	bech, err := bech32.Encode(hrp, combined)
+	if err != nil {
+		return "", err
+	}
+
+	// Check validity by decoding the created address.
+	version, program, err := decodeSegWitAddress(bech)
+	if err != nil {
+		return "", fmt.Errorf("invalid segwit address: %v", err)
+	}
+
+	if version != witnessVersion || !bytes.Equal(program, witnessProgram) {
+		return "", fmt.Errorf("invalid segwit address")
+	}
+
+	return bech, nil
 }
 
 // Address is an interface type for any type of destination a transaction
@@ -81,6 +134,40 @@ type Address interface {
 // When the address does not encode the network, such as in the case of a raw
 // public key, the address will be associated with the passed defaultNet.
 func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
+	// Bech32 encoded segwit addresses start with a human-readable part
+	// (hrp) followed by '1'. For Bitcoin mainnet the hrp is "bc", and for
+	// testnet it is "tb". If the address string has a prefix that matches
+	// one of the prefixes for the known networks, we try to decode it as
+	// a segwit address.
+	oneIndex := strings.LastIndexByte(addr, '1')
+	if oneIndex > 1 {
+		prefix := addr[:oneIndex+1]
+		if chaincfg.IsBech32SegwitPrefix(prefix) {
+			witnessVer, witnessProg, err := decodeSegWitAddress(addr)
+			if err != nil {
+				return nil, err
+			}
+
+			// We currently only support P2WPKH and P2WSH, which is
+			// witness version 0.
+			if witnessVer != 0 {
+				return nil, UnsupportedWitnessVerError(witnessVer)
+			}
+
+			// The HRP is everything before the found '1'.
+			hrp := prefix[:len(prefix)-1]
+
+			switch len(witnessProg) {
+			case 20:
+				return newAddressWitnessPubKeyHash(hrp, witnessProg)
+			case 32:
+				return newAddressWitnessScriptHash(hrp, witnessProg)
+			default:
+				return nil, UnsupportedWitnessProgLenError(len(witnessProg))
+			}
+		}
+	}
+
 	// Serialized public keys are either 65 bytes (130 hex chars) if
 	// uncompressed/hybrid or 33 bytes (66 hex chars) if compressed.
 	if len(addr) == 130 || len(addr) == 66 {
@@ -117,6 +204,49 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
 	default:
 		return nil, errors.New("decoded address is of unknown size")
 	}
+}
+
+// decodeSegWitAddress parses a bech32 encoded segwit address string and
+// returns the witness version and witness program byte representation.
+func decodeSegWitAddress(address string) (byte, []byte, error) {
+	// Decode the bech32 encoded address.
+	_, data, err := bech32.Decode(address)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	// The first byte of the decoded address is the witness version, it must
+	// exist.
+	if len(data) < 1 {
+		return 0, nil, fmt.Errorf("no witness version")
+	}
+
+	// ...and be <= 16.
+	version := data[0]
+	if version > 16 {
+		return 0, nil, fmt.Errorf("invalid witness version: %v", version)
+	}
+
+	// The remaining characters of the address returned are grouped into
+	// words of 5 bits. In order to restore the original witness program
+	// bytes, we'll need to regroup into 8 bit words.
+	regrouped, err := bech32.ConvertBits(data[1:], 5, 8, false)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	// The regrouped data must be between 2 and 40 bytes.
+	if len(regrouped) < 2 || len(regrouped) > 40 {
+		return 0, nil, fmt.Errorf("invalid data length")
+	}
+
+	// For witness version 0, address MUST be exactly 20 or 32 bytes.
+	if version == 0 && len(regrouped) != 20 && len(regrouped) != 32 {
+		return 0, nil, fmt.Errorf("invalid data length for witness "+
+			"version 0: %v", len(regrouped))
+	}
+
+	return version, regrouped, nil
 }
 
 // AddressPubKeyHash is an Address for a pay-to-pubkey-hash (P2PKH)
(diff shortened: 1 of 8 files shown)


halseth:
Will need https://github.com/btcsuite/btcd/pull/959 to build.

address.go
```diff
@@ -44,6 +48,39 @@ func encodeAddress(hash160 []byte, netID byte) string {
...
+	witnessProgram []byte) (string, error) {
+
+	// Group the address bytes into 5 bit groups, as this is what is used to
+	// encode each character in the address string.
+	converted, err := bech32.ConvertBits(witnessProgram, 8, 5, true)
+	if err != nil {
+		return "", err
+	}
+
+	// Concatenate the witness version and program, and encode the resulting bytes
```
Roasbeef: Within the project, lines are typically wrapped to 80 characters on a best-effort basis. This line in particular exceeds this limit (and a few others as well) as it's 86 characters long.

address.go
```diff
@@ -81,6 +118,22 @@ type Address interface {
 // When the address does not encode the network, such as in the case of a raw
 // public key, the address will be associated with the passed defaultNet.
 func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
+
+	// Bech32 encoded segwit addresses start with an human-readable part  (hrp)
```
Roasbeef: There's an extra space after "part".

address.go
```diff
@@ -81,6 +118,22 @@ type Address interface {
...
+	// "tb". If the address string matches the hrp found in the defaultNet params,
+	// we try to decode it as a segwit address.
+	hrp := defaultNet.Bech32HRPSegwit
+	prefix := hrp + "1"
+	if strings.HasPrefix(strings.ToLower(addr), prefix) {
+		witnessVer, witnessProg, err := decodeSegWitAddress(hrp, addr)
+		if err != nil {
+			return nil, err
+		}
+		// Successfully parsed addr as segwit address
```
Roasbeef: Missing a period at the end of this line. Also a new line should be inserted above.

address.go
```diff
@@ -119,6 +172,61 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
 	}
 }
 
+// decodeSegWitAddress parses a bech32 encoded segwit address string and
+// returns the witness version and witness program byte representation.
+func decodeSegWitAddress(hrp, address string) (byte, []byte, error) {
+	// We MUST verify that the human-readable part is "bc" for mainnet and "tb"
+	// for testnet.
+	hrp = strings.ToLower(hrp)
+	if hrp != "bc" && hrp != "tb" {
```
Roasbeef: This'll trivially break in the case that additional prefixes are added. For example, with this, one isn't able to re-use this library for Litecoin or any other coins that've adopted `bech32`. 

To fix this, the `chaincfg.Params` should be passed in and the `hrp` extracted compared against the defined `Bech32HRPSegwit` for that network.

address.go
```diff
@@ -119,6 +172,61 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
...
+		return 0, nil, fmt.Errorf("no witness version")
+	}
+
+	// ...and be <= 16.
+	version := data[0]
+	if version > 16 {
+		return 0, nil, fmt.Errorf("invalid witness version: %v", version)
+	}
+
+	// The remainding characters of the address is each decoded using 5 bits.
```
Roasbeef: remainding -> remaining 

is -> are

Roasbeef: Minor re-write of the comment: 
> The remaining characters of the address returned are grouped into words of 5 bits. In order to restore the original witness program bytes, we'll need to regroup into 8 bit words.

address.go
```diff
@@ -119,6 +172,61 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
...
+	// We must regroup these to regular 8 bit groups.
+	regrouped, err := bech32.ConvertBits(data[1:], 5, 8, false)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	// The regrouped data must be between 2 and 40 bytes
```
Roasbeef: Missing a period.

address.go
```diff
@@ -119,6 +172,61 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
...
+	if len(regrouped) < 2 || len(regrouped) > 40 {
+		return 0, nil, fmt.Errorf("invalid data length")
+	}
+
+	// For witness version 0, address MUST be exactly 20 or 32 bytes
```
Roasbeef: Missing a period.

address.go
```diff
@@ -375,3 +483,82 @@ func (a *AddressPubKey) AddressPubKeyHash() *AddressPubKeyHash {
 func (a *AddressPubKey) PubKey() *btcec.PublicKey {
 	return a.pubKey
 }
+
+// AddressSegwit is an Address for a pay-to-witness-pubkey-hash (P2WPKH) or
```
Roasbeef: These should be split up into `AddressWitnessPubKeyHash` and `AddressWitnessScriptHash` in order to properly mirror the aspects of the prior PR that this one is replacing. With these changes, it should be able to slip in without any modifications to my prior PR.

address.go
```diff
@@ -375,3 +483,82 @@ func (a *AddressPubKey) AddressPubKeyHash() *AddressPubKeyHash {
...
+	net *chaincfg.Params) (*AddressSegwit, error) {
+	return newAddressSegwit(net.Bech32HRPSegwit, witnessVer, witnessProg)
+}
+
+// newAddressSegwit is an internal helper function to create a AddressSegwit
+// with a known human-readable part, rather than looking it up through its
+// parameters.
+func newAddressSegwit(hrp string, witnessVer byte,
+	witnessProg []byte) (*AddressSegwit, error) {
+	// Check for valid program length for witness version 0
```
Roasbeef: Missing a newline above.

bech32/bech32.go
```diff
@@ -0,0 +1,211 @@
...
+	if len(bech) > 90 {
+		return "", nil, fmt.Errorf("string too long")
+	}
+
+	// The human-readable part is everything before the last '1'
+	hrp := bech[:one]
+	data := bech[one+1:]
+
+	// Each character corresponds to the byte with value of the index in 'charset'
+	var decoded []byte
```
Roasbeef: This slice can be pre-allocated as so: 
```go
decoded := make([]byte, len(data))
```

bech32/bech32.go
```diff
@@ -0,0 +1,211 @@
...
+// (base32).
+func Encode(hrp string, data []byte) (string, error) {
+	// Calculate the checksum of the data and append it at the end.
+	checksum := bech32Checksum(hrp, data)
+	combined := append(data, checksum...)
+
+	// The resulting bech32 string is the concatenation of the hrp, the separator
+	// 1, data and checksum. Everything after the separator is represented using
+	// the specified charset.
+	result := []byte(hrp)
```
Roasbeef: This can can also be pre-allocated: 
```go
result := make([]byte, len(hrp)+1+len(combined))
```

bech32/bech32.go
```diff
@@ -0,0 +1,211 @@
...
+	for _, b := range combined {
+		if int(b) >= len(charset) {
+			return "", fmt.Errorf("invalid data byte: %v", b)
+		}
+		result = append(result, charset[b])
+	}
+	return string(result), nil
+}
+
+// ConvertBits converts a byte slice where each byte is encoding fromBits bits,
```
Roasbeef: This is a nice utility function, but from a user's point of view, I'd expect it to be called from within the `Encode` function. As it now, users _must_ manually call this function, otherwise they won't properly be encoding in the `bech32` format.

bech32/bech32.go
```diff
@@ -0,0 +1,211 @@
...
+// to a byte slice where each byte is encoding toBits bits.
+func ConvertBits(data []byte, fromBits, toBits uint8, pad bool) ([]byte, error) {
+	if fromBits < 1 || fromBits > 8 || toBits < 1 || toBits > 8 {
+		return nil, fmt.Errorf("only bit groups between 1 and 8 allowed")
+	}
+
+	// The final bytes, each byte encoding toBits bits
```
Roasbeef: All comments should end in a period (applies to the rest of this file).

bech32/bech32.go
```diff
@@ -0,0 +1,211 @@
...
+	// out of the toBits goal.
+	nextByte := byte(0)
+	filledBits := uint8(0)
+
+	for _, b := range data {
+
+		// discard unused bits
+		b = b << (8 - fromBits)
+
+		// How many bits remainding to extract from the input data
```
Roasbeef: remainding -> remaining 

The above error occurs a few times in this file.

bech32/bech32.go
```diff
@@ -0,0 +1,211 @@
...
+	if filledBits > 0 && (filledBits > 4 || nextByte != 0) {
+		return nil, fmt.Errorf("invalid incomplete group")
+	}
+
+	return regrouped, nil
+}
+
+// For more on details on the checksum calculation, please refer to BIP 173.
+func bech32Checksum(hrp string, data []byte) []byte {
+	// Convert the bytes to list of integers, as this is needed fot the checksum
```
Roasbeef: fot -> for

address.go
halseth: I originally did this, but realized that these two would restrict us to version `0` address types (I might be wrong here, so correct me if I am). In case we encounter a non-version 0 address, we _could_ fail parsing it, but that would make the implementation non-forwards compatible. We could also create a third type to hold non-zero types, but maybe that is unnecessary complex?

bech32/bech32.go
halseth: I totally agree with this, but I left the bit conversion outside the Bech32 `Encode` method to mimic the reference implementation: https://github.com/sipa/bech32/blob/master/ref/python/segwit_addr.py#L120

BIP173 also seem to indicate that the bit conversion should be done _before_ Bech32 encoding: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#segwit-address-format, but this is not entirely clear to me.

address.go
Roasbeef: `p2wkh` and `p2wsh` are _only_ defined for version zero witness programs. I don't think we need to worry about the case of non-zero for now, as in that event a new BIP would need to be created in order to reach wide adoption.

address.go
Roasbeef: LGTM

bech32/bech32.go
Roasbeef: Hmm, yeh I can see it both ways really. 

Upon reflecting I think it's fine as is, we'll be re-using this package (and that method) within `lnd` very soon in order to implement the payment request format.

address.go
halseth: Makes sense. Done.

halseth:
Done 👍

address.go
```diff
@@ -81,6 +118,37 @@ type Address interface {
...
 // public key, the address will be associated with the passed defaultNet.
 func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
+
+	// Bech32 encoded segwit addresses start with an human-readable part
+	// (hrp) followed by 1. For Bitcoin mainnet the hrp is "bc", and for
+	// testnet it is "tb". If the address string matches the hrp found in
+	// the defaultNet params, we try to decode it as a segwit address.
+	hrp := defaultNet.Bech32HRPSegwit
+	prefix := hrp + "1"
+	if strings.HasPrefix(strings.ToLower(addr), prefix) {
```
davecgh: This section is incorrect.  It is an error to assume the caller is passing the current network.  The default network (hence the name `defaultNet`) only applies to addresses that do not have a network encoded with them (the only one that is true for currently is raw pay-to-pubkey addresses).

A bech32 address, by definition, must have an hrp that identifies the network.  See the code below which calls `chaincfg.IsPubKeyHashAddrID(netID)` to see how this is handled for old style addresses.  The same approach must be followed here since the network has to be decoded from the string and it must be done according to registered networks via the chain parameters.

address.go
```diff
@@ -81,6 +118,37 @@ type Address interface {
 // When the address does not encode the network, such as in the case of a raw
 // public key, the address will be associated with the passed defaultNet.
 func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
+
```
davecgh: Extraneous newline.

address.go
```diff
@@ -119,6 +187,57 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
 	}
 }
 
+// decodeSegWitAddress parses a bech32 encoded segwit address string and
+// returns the witness version and witness program byte representation.
+func decodeSegWitAddress(hrp, address string) (byte, []byte, error) {
+
```
davecgh: Extraneous newline.

address.go
```diff
@@ -44,6 +48,39 @@ func encodeAddress(hash160 []byte, netID byte) string {
 	return base58.CheckEncode(hash160[:ripemd160.Size], netID)
 }
 
+// encodeSegWitAddress creates a bech32 encoded address string representation
+// from witness version and witness program.
+func encodeSegWitAddress(hrp string, witnessVersion byte,
+	witnessProgram []byte) (string, error) {
```
davecgh: Please keep function definitions on the same line to be more consistent with the rest of code base.

address.go
```diff
@@ -81,6 +118,37 @@ type Address interface {
 // When the address does not encode the network, such as in the case of a raw
 // public key, the address will be associated with the passed defaultNet.
 func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
+
+	// Bech32 encoded segwit addresses start with an human-readable part
+	// (hrp) followed by 1. For Bitcoin mainnet the hrp is "bc", and for
+	// testnet it is "tb". If the address string matches the hrp found in
+	// the defaultNet params, we try to decode it as a segwit address.
+	hrp := defaultNet.Bech32HRPSegwit
+	prefix := hrp + "1"
```
davecgh: This should be part of the chaincfg parameters since it's always the case.

bech32/README.md
```diff
@@ -0,0 +1,34 @@
...
+[![Build Status](http://img.shields.io/travis/btcsuite/btcutil.svg)]
+(https://travis-ci.org/btcsuite/btcutil) [![ISC License]
+(http://img.shields.io/badge/license-ISC-blue.svg)](http://copyfree.org)
+[![GoDoc](https://godoc.org/github.com/btcsuite/btcutil/bech32?status.png)]
+(http://godoc.org/github.com/btcsuite/btcutil/bech32)
+
+Package bech32 provides a Go implementation of the bech32 format specified in
+[BIP 173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki).
+
+Test vectors from BIP 173 is added to ensure compatibility with the reference
```
davecgh: s/is/are/

davecgh: s/reference implementation/BIP/

bech32/bech32_test.go
```diff
@@ -0,0 +1,62 @@
...
+package bech32_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/btcsuite/btcutil/bech32"
+)
+
+var validBechStrings = []string{
```
davecgh: Please move these definitions inside of the test function scope as to avoid polluting the package namespace.

Ideally, they would be converted to table-driven tests to match almost all the rests of the tests in the code base along the lines of the following, however, I'm fine with the being separate variables so long as they are defined inside the test function scope.

```Go

func TestBech32(t *testing.T) {
	tests := struct{
		str string
		valid bool
	}{
		{"A12UEL5L", true},
		// ...
		{"split1cheo2y9e2w", false},
	}
	for _, test := range tests {
		hrp, decoded, err := bech32.Decode(test.str)
		if test.valid && err != nil {
				t.Errorf("decode of valid string returned unexpected error %v", err)
				continue
		}

		// ....
	}
}

File: bech32/cov_report.sh [hunk 23]
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
davecgh: Please remove this file.  We really should remove it from the other packages as well.  It is no longer supported since `go test` provides coverage nowadays.

address.go
```diff
@@ -44,6 +48,39 @@ func encodeAddress(hash160 []byte, netID byte) string {
...
+	// Concatenate the witness version and program, and encode the resulting
+	// bytes using bech32 encoding.
+	combined := append([]byte{witnessVersion}, converted...)
+	bech, err := bech32.Encode(hrp, combined)
+	if err != nil {
+		return "", err
+	}
+
+	// Check validity by decoding the created address.
+	version, program, err := decodeSegWitAddress(hrp, bech)
```
davecgh: I don't think this should be necessary.  The `bech32` package is already proven correct by its tests.  Adding the overhead to ensure it is doing the job it is designed to do seems wasteful.

address.go
```diff
@@ -81,6 +118,37 @@ type Address interface {
...
+	if strings.HasPrefix(strings.ToLower(addr), prefix) {
+		witnessVer, witnessProg, err := decodeSegWitAddress(hrp, addr)
+		if err != nil {
+			return nil, err
+		}
+
+		// We currently only support P2WPKH and P2WSH, which is
+		// witness version 0.
+		if witnessVer != 0 {
+			return nil, fmt.Errorf("unsupported witness "+
```
davecgh: Please define programmatically detectable errors for these as the rest of the function does.

```Go
// Comments...
type UnsupportedWitnessVerError byte
func (e UnsupportedWitnessVerError) Error() string {
	return "unsupported witness version: " + string(e)
}


...
...
	if witnessVer != 0 {
		return nil, UnsupportedWitnessVerError(witnessVer)
	}
```

address.go
```diff
@@ -81,6 +118,37 @@ type Address interface {
...
+				"version: %v", witnessVer)
+		}
+
+		switch len(witnessProg) {
+		case 20:
+			return newAddressWitnessPubKeyHash(hrp, witnessProg)
+		case 32:
+			return newAddressWitnessScriptHash(hrp, witnessProg)
+		default:
+			return nil, fmt.Errorf("unsupported witness program "+
```
davecgh: Same here regarding a programmatically detectable error.

halseth: Added `UnsupportedWitnessProgLenError` type.

address.go
halseth: Added the type to the top of the file, not really sure where it belongs.

address.go
davecgh: Correct.  Although you'll also of course need to make sure you're not using an invalid index as the above would do.  It would panic in the case of a malformed address.

```
	// The network for a bech32 is identified by a human-readable part followed
	// by a 1.
	oneIndex := strings.LastIndexByte(addr, '1')
	if oneIndex != -1 {
			// ....
	}
```

halseth: I added the utility method in https://github.com/btcsuite/btcd/pull/968, and updated the code here to use it.

address.go
halseth: I see what you're getting at, I didn't really consider this method to be used for non-bitcoin addresses. If we were to make this chain agnostic, then I guess we would move Bitcoin specific code/verification into `params.chaincfg`? I'm not really sure how that would work, but would be happy to do it if I could get some guidance :)

address.go
```diff
@@ -81,6 +135,40 @@ type Address interface {
 // When the address does not encode the network, such as in the case of a raw
 // public key, the address will be associated with the passed defaultNet.
 func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
+	// Bech32 encoded segwit addresses start with a human-readable part
+	// (hrp) followed by '1'. For Bitcoin mainnet the hrp is "bc", and for
+	// testnet it is "tb". If the address string has a prefix that matches
+	// one of the prefixes for the known networks, we try to decode it as
+	// a segwit address.
+	oneIndex := strings.LastIndexByte(addr, '1')
+	if oneIndex != -1 {
```
halseth: I think tou are right about that, the BIP doesn't specify it, but the reference implementation does the same check, so I made the `bech32.Decode` method require a non-empty hrp. Will change to `if oneIndex > 0`

address.go
```diff
@@ -119,6 +201,56 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
 	}
 }
 
+// decodeSegWitAddress parses a bech32 encoded segwit address string and
+// returns the witness version and witness program byte representation.
+func decodeSegWitAddress(hrp, address string) (byte, []byte, error) {
```
halseth: my bad, after introducing the `IsBech32SegwitPrefix ` check, this should no longer be necessary.

bech32/bech32.go
halseth: Again, I did this to mimic the reference implementation, but if we all agree we should move it inside `Encode` I can do that? :)

halseth: After I started implementing this change, I noticed some problems. I have been thinking a bit about this, and I now think that it does indeed make more sense to do it the way it is done currently (and this is probably the reason it is done this way in the reference implementation). Since the first 5 bits are used to encode the witness version, moving the bit conversion inside the `bech32.Encode` method would lead to some unnecessary complexities IMO: 

1. Since `bech32` is used to encode arbitrary data, in this case we would need to create a byte array where the first 5 bits are the witness version, the next 3 bits would be the first 3 bits of the witness program and so on (everything shifted). It might not even align properly and would need to be padded. This would kind of defeat the purpose of moving the `ConvertBits` logic inside `bech32.Encode`, as we would need similar logic outside anyway.
2. When decoding a bech32 string, `bech32.Decode` would use `ConvertBits` to create and return a proper byte array (not base32 as now). Again, we would need to interpret the first 5 bits as the witness version, then shift the remaining bytes to get the witness program.

Reading the BIP from the point of view that “Bech32” is a way of encoding base32 data, it is more clear. This is why the “Segwit address format” section specifies that the data should be regrouped to 5 bit groups before encoding it using bech32.

Roasbeef: I've shifted a bit, and I don't consider this to be a strong blocker for merging in this PR. The entire need to do this 8-bit to 5-bit word conversion and back is a quirk of `bech32` itself. As @halseth points out, this quirk is repeated in the reference implementation, and all derived from it. In the future, once we have a bit more client usage, if we have strong feedback one way or the other, we can introduce the change.

bech32/bech32.go
```diff
@@ -0,0 +1,220 @@
...
+	return hrp, decoded[:len(decoded)-6], nil
+}
+
+// Encode encodes a byte slice into a bech32 string with the
+// human-readable part hrb. Note that the bytes must each encode 5 bits
+// (base32).
+func Encode(hrp string, data []byte) (string, error) {
+	// Calculate the checksum of the data and append it at the end.
+	checksum := bech32Checksum(hrp, data)
+	combined := append(data, checksum...)
```
davecgh: I would prefer this not be done here.  It's an extra unnecessary allocation.  Just append the data and checksum directly to the result, which is already properly sized below and replace the encoded characters directly in place.

e.g.

```Go
dataOffset := len(hrp)+1
checksum := bech32Checksum(hrp, data)
result := make([]byte, 0, len(hrp)+1+len(data)+len(checksum))
result = append(result, hrp...)
result = append(result, '1')
result = append(result, data...)
result = append(result, checksum...)
for i, b := range result[dataOffset:] {
		if int(b) >= len(charset) {
			fmt.Printf("invalid data byte: %v", b)
			return
		}
		result[dataOffset+i] = charset[b]
}
```

bech32/bech32.go
davecgh: That seems like a fair approach.  I agree that should not hold up this PR.

bech32/bech32.go
```diff
@@ -0,0 +1,220 @@
...
+const charset = "qpzry9x8gf2tvdw0s3jn54khce6mua7l"
+
+var gen = []int{0x3b6a57b2, 0x26508e6d, 0x1ea119fa, 0x3d4233dd, 0x2a1462b3}
+
+// Decode decodes a bech32 encoded string, returning the human-readable
+// part and the data part excluding the checksum.
+func Decode(bech string) (string, []byte, error) {
+	// Only	ASCII characters between 33 and 126 are allowed.
+	for i := 0; i < len(bech); i++ {
+		if bech[i] < 31 || bech[i] > 126 {
```
halseth: wow, yeah my bad. Fixed and added tests.

bech32/bech32.go
halseth: rewrote this a bit. PTAL.

davecgh:
Thanks for the update.  This looks good now.  Please go ahead and squash it, then I'll approve and merge it.

halseth:
Thanks! Done :)

PR #102, opened 2017-06-11
Expose hdkeychain.Depth() 

afk11:
which returns the number of derivations that have occurred since the master. Added tests of the function to existing test cases

### DIFF
--- a/hdkeychain/extendedkey_test.go
+++ b/hdkeychain/extendedkey_test.go
@@ -231,6 +231,11 @@ tests:
 			}
 		}
 
+		if extKey.Depth() != uint8(len(test.path)) {
+			t.Errorf("Depth of key %d should match fixture path")
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


hdkeychain/extendedkey_test.go
```diff
@@ -1042,6 +1047,9 @@ func TestMaximumDepth(t *testing.T) {
 	}
 
 	for i := uint8(0); i < math.MaxUint8; i++ {
+		if extKey.Depth() != i {
+			t.Fatal("extendedkey depth %d should match expected value %d", extKey.Depth(), i)
```
davecgh: Please keep to 80 cols to be consistent with the rest of the code.

Also, it has to be `t.Fatalf` since it's a format string.

afk11: Ah you're quite right - fixed

PR #114, opened 2018-01-31
Fix error in base58check test

stevenroose:
Currently the test checks the same empty string in every iteration of the loop.


jcvernaleo:
Close in favor of the rebased #174

PR #121, opened 2018-07-06
gcs/builder: remove the AddScript method as it's no longer used

Roasbeef:
In this commit, we remove the AddScript method as it's no longer used,
and AddEntry should be used in place for adding pkScripts to the
filters.


wpaulino:
LGTM

PR #124, opened 2018-08-10
filter: add Size method

rawtxapp:
Context: https://github.com/lightninglabs/neutrino/pull/72

Filter implements Size which will be used by Cache inside neutrino.

### DIFF
--- a/gcs/gcs.go
+++ b/gcs/gcs.go
@@ -430,3 +430,8 @@ func (f *Filter) readFullUint64(b *bstream.BStream) (uint64, error) {
 	v := (quotient << f.p) + remainder
 	return v, nil
 }
+
+// Size returns the Filter's size in bytes.
+func (f *Filter) Size() uint64 {
+	return uint64(len(f.filterData))
+}


gcs/gcs.go
```diff
@@ -430,3 +430,8 @@ func (f *Filter) readFullUint64(b *bstream.BStream) (uint64, error) {
 	v := (quotient << f.p) + remainder
 	return v, nil
 }
+
+// Size returns the Filter's size in bytes.
+func (f *Filter) Size() uint64 {
+	return uint64(len(f.filterData))
```
Roasbeef: This is wrong as it should include the size serialized. Why can't `NBytes` be used for this?

rawtxapp: My understanding was that filterData is what's actually consuming the memory since it's the one the struct holds.

So you suggest doing `len(f.NBytes())` ?

rawtxapp:
Closing this for now, will wrap the filter in neutrino as suggested by Johan in the context PR.

PR #125, opened 2018-08-26
Include empty scripts in basic block filter.

jimpo:


jimpo:
Decided against this.

PR #126, opened 2018-08-31
PSBT BIP 174 implementation

AdamISZ:
See next.


AdamISZ:
An update to the above: I believe the implementation is mostly complete now, at least in as far as I see immediate uses. What is not present:

* No implementation of the Combiner role as of yet; but in the cases where this is applicable, one can simply apply the Updater role multiple times
* In contrast with the previous comment, I've now removed any signing interface, which probably just adds an unnecessary layer; thus, a caller/user of this API simply uses a call to `PsbtUpdater.AddPartialSignature`, and this function implements the important checks laid out in the BIP on signature validity w.r.t the transaction it's signing (i.e. the Signer role is included within that function).
* The validation checks in finalizing (which constructs final scriptSig and witness fields from the data fields) assumes that p2sh non-witness is multisig, and p2wsh is multisig (though it allows any M of N). Generalizing that scriptSig/witness construction to any custom script is an obviously desirable improvement.
* Use of BIP32 derivation fields is still an open question; they are serialized and can be added with `PsbtUpdater.AddInBip32Derivation`, but there is no use of them in validation here, I presume that in some cases they will be used by callers (such as hardware wallets, other constrained environments) to re-derive signing keys / check them.

It is generally functional however, and I'm "dogfooding" it a bit by using it in the CoinJoin rpc implementation in my fork [here](https://github.com/AdamISZ/lnd/tree/coinjoin-expts).

I'll now add several more tests to `psbt_tests.go` to check correct functioning for a broader set of input types, particularly.

afk11:
Nice one, this looks good! Will try give this PR a review soon

psbt/extractor.go
```diff
@@ -0,0 +1,80 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			// Then for witCount number of stack items, each item
+			// has a varint length prefix, followed by the witness
+			// item itself.
+			tin.Witness = make([][]byte, witCount)
+			for j := uint64(0); j < witCount; j++ {
+				count, err := wire.ReadVarInt(r, 0)
```
afk11: This section can be extracted out as wire.ReadVarBytes

psbt/finalizer.go
```diff
@@ -0,0 +1,497 @@
...
+// information to finalize this input, based on the template
+// implied by addrType (currently only 'p2wpkh, 'p2wsh', 'np2wpkh' or
+// 'np2wsh' are accepted, where 'n' refers to nested).
+// TODO Use something from the lib (not string) for address type,
+// and implement the code for other address types.
+func isFinalizable(p *Psbt, idx int, addrType string) bool {
+	iKVPLs := *(p.InKVPairs)
+	iKVPL := iKVPLs[idx]
+	switch addrType {
+	case "p2wpkh":
```
afk11: The named script-hash script types feels wrong.. what about passing around (isP2sh, isP2wsh, script)? Then the code you write is more generalizable towards [, P2SH, P2WSH, P2SH|P2WSH] handling. (you don't want to write multisig support for each script-hash type or for each m-of-n for that matter)

afk11:
Why parse into PsbtKeyValuePairs instead of structures containing the actual types? (wire.Tx, wire.TxOut, etc). Less change of deserializing a lot of garbage if you can do sanity checks while parsing.

psbt/psbt.go
```diff
@@ -0,0 +1,676 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// validatePubkey returns true if the passed data is a valid
+// compressed serialization of a valid secp256k1 pubkey.
+// It is a requirement of BIP174 that only compressed pubkeys
+// are allowed.
```
afk11: I'm not sure this last point is correct - I can't see this anywhere in the BIP, and the reference implementation allows for uncompressed keys

psbt/psbt.go
```diff
@@ -0,0 +1,676 @@
...
+}
+
+// SerializeBIP32Derivation takes a master key fingerprint
+// as defined in BIP32, along with a path specified as a list
+// of uint32 values, and returns a bytestring specifying the derivation
+// in the format required by BIP174:
+// // master key fingerprint (4) || child index (4) || child index (4) || ...
+func SerializeBIP32Derivation(masterKeyFingerprint []byte,
+	bip32Path []uint32) []byte {
+	//TODO fixed length of 4 bytes
```
afk11: This comment is about the masterKeyFingerprint being variable size? You could [4]byte it, I'd personally use uint32 for this

psbt/psbt.go
```diff
@@ -0,0 +1,676 @@
...
+var (
+	PsbtMagic = [PsbtMagicLength]byte{0x70,
+		0x73, 0x62, 0x74, 0xff, // = "psbt" + 0xff sep
+	}
+)
+
+// The below are the known key types as per the BIP.
+// Unknown types may be accepted but will not be processed.
+const (
+
```
afk11: Hmm, guessing these are the same type because they're used in the PsbtKeyValuePair{}.. I'd probably make them distinct types though, and use GlobalType, InputType, OutputType..

psbt/psbt.go
AdamISZ: Wow, I've no idea what happened here; somehow I had it in my head that this requirement was specified in the BIP, but there nothing saying that there now I read it again ...

It is a requirement of BIP141, I *guess* I may have confused myself by mixing that up, but weird, in any case. Will fix, thanks.

psbt/finalizer.go
AdamISZ: Yes, as noted in the comment just above, this is definitely inappropriate and should be replaced with something in-lib.

AdamISZ:
> Why parse into PsbtKeyValuePairs instead of structures containing the actual types? (wire.Tx, wire.TxOut, etc). Less chance of deserializing a lot of garbage if you can do sanity checks while parsing

I feel like I could interpret this a couple of different ways; one thing is, it probably makes a lot of sense to have a struct for PsbtInput (I note that the Core implementation does this), rather than having it be a list of kv-pairs. Not that I think the difference would be that huge, but it should probably be somewhat cleaner. I definitely considered it.

On the other hand, I think you're actually asking about the idea of having separate structs for each k-v type (like a struct for PsbtInWitnessUtxo, PsbtInNonWitnessUtxo, ... etc.). I actually didn't consider that; it sounds like a sound idea, not sure whether I'd get a bit tripped up with not having in-built inheritance (golang doesn't do inheritance in quite the same simple way as other langs.).

psbt/extractor.go
AdamISZ: Ah, thanks, good catch!

afk11:
> I feel like I could interpret this a couple of different ways; one thing is, it probably makes a lot of sense to have a struct for PsbtInput

Yes, this is what I meant mainly! Works with separate types for GlobalKeyType+PSBT global / InputKeyType + PSBT input / etc

> On the other hand, I think you're actually asking about the idea of having separate structs for each k-v type (like a struct for PsbtInWitnessUtxo, PsbtInNonWitnessUtxo, ... etc.)

Hmm, might be useful for the Unknown / Derivations / PartialSig fields since they're a map type.

psbt/extractor.go
AdamISZ: done: 2d9dc6ae54f887d6cf29aa6c6e46eacfaa82d7de

psbt/psbt.go
AdamISZ: Yes, I went with your suggestion of using uint32, but only because it looks ugly to copy a byte slice to a byte array. Not that it really matters I guess. Done here: b259236475cd8705a63e330a8904070036a349d7

psbt/psbt.go
AdamISZ: Fixed this (check for any valid pubkey) in abf27ad43f22f7556bf9b0c4e7bf5e7b3cb8ea03

Also validates signatures.

AdamISZ:
Just dropping a quick note here, in case anyone reads, that I'm in the process of updating much as per @afk11 's comments (separating out the serialization from the logical manipulation using structs for PsbtInput etc.). Expect to post a big refactoring of this here in the next few days, probably.

AdamISZ:
Finished above-mentioned refactoring.

There's no change to the description as per [this comment](https://github.com/btcsuite/btcutil/pull/126#issuecomment-421555971).

AdamISZ:
First, thanks a lot for taking a look.
Second: I'm trying to replicate your error with the given raw tx, using `NewPsbtFromUnsignedTx` . Running this:

```go
serTx, err := hex.DecodeString("00000000000101e165f072311e71825b47a4797221d7ae56d4b40b7707c540049aee43302448a40000000000feffffff0212f1126a0000000017a9143e836801b2b15aa193449d815c62d6c4b6227c898780778e060000000017a914ba4bdb0b07d67bc60f59c1f4fe54170565254974870000000000")
	if err != nil {
		t.Fatalf("Error: %v", err)
	}
	tx := wire.NewMsgTx(2)
	err = tx.Deserialize(bytes.NewReader(serTx))
	if err != nil {
		t.Fatalf("Error: %v", err)
	}
	psbt1, err := NewPsbtFromUnsignedTx(tx)
	if err != nil {
		t.Fatalf("Error: %v", err)
	}
	fmt.Println(psbt1.B64Encode())
```
fails on the call to create `psbt1` with "Invalid PSBT, raw transaction must be unsigned". So this raises two things:

(1) should I have code here that adds the ability to load a partially signed *bitcoin serialized* (not BIP174 serialized) transaction, or should it remain with the insistence that there are no signatures.

(2) How is it that the call to `NewPsbtFromUnsignedTx` actually worked for you and output a base64 encoded string?

Will keep looking at this and also address your design-related comments, which I suspect are quite right.

AdamISZ:
Yes, it was a bug. On line 717 of psbt.go, replace `tin.Witness != nil` with `len(tin.Witness) != 0` and it replicates the same base64 as Core.

(Bug was that I forgot `tin.Witness` is a slice, empty doesn't equal nil).

Will push fix once I've thought through the other issues you brought up.

AdamISZ:
@RHavar re your second comment:

Yes, the `Raw` element is unnecessary, I agree.

Memo-ization as you say could be a justification, but def not worth it here, it's quite hard to imagine a situation where perf. would ever be an issue. Doesn't justify adding unnecessary code/state.

Re: pointer to slice instead of just slice, I also agree, that looks like a bit of a derp to be honest :)

RHavar:
> and it replicates the same base64 as Core.

Does it? Here's my full test file:

```
package main

import (
	"github.com/btcsuite/btcd/wire"
	"bytes"
	"fmt"
	"encoding/hex"
	"log"
	"github.com/btcsuite/btcutil/psbt"
)

func main() {
	serTx, err := hex.DecodeString("00000000000101e165f072311e71825b47a4797221d7ae56d4b40b7707c540049aee43302448a40000000000feffffff0212f1126a0000000017a9143e836801b2b15aa193449d815c62d6c4b6227c898780778e060000000017a914ba4bdb0b07d67bc60f59c1f4fe54170565254974870000000000")
	if err != nil {
		log.Fatalf("Error: %v", err)
	}
	tx := wire.NewMsgTx(2)
	err = tx.Deserialize(bytes.NewReader(serTx))
	if err != nil {
		log.Fatalf("Error: %v", err)
	}
	psbt1, err := psbt.NewPsbtFromUnsignedTx(tx)
	if err != nil {
		log.Fatalf("Error: %v", err)
	}
	fmt.Println(psbt1.B64Encode())
}
```

And it prints:

```
cHNidP8BAHYAAAAAAAEB4WXwcjEecYJbR6R5ciHXrlbUtAt3B8VABJruQzAkSKQAAAAAAP7///8CEvESagAAAAAXqRQ+g2gBsrFaoZNEnYFcYtbEtiJ8iYeAd44GAAAAABepFLpL2wsH1nvGD1nB9P5UFwVlJUl0hwAAAAAAAAAAAA==
```

which not equiv to what core gives (it gives me)

```
cHNidP8BAHMAAAAAAeFl8HIxHnGCW0ekeXIh165W1LQLdwfFQASa7kMwJEikAAAAAAD+////AhLxEmoAAAAAF6kUPoNoAbKxWqGTRJ2BXGLWxLYifImHgHeOBgAAAAAXqRS6S9sLB9Z7xg9ZwfT+VBcFZSVJdIcAAAAAAAAAAA==
```

And if i run `bitcoin-cli decodepsbt` it works with the output of core, but fails with the output of this PR

AdamISZ:
Did you apply the patch? I tested it as giving exactly that output after the patch.

RHavar:
> Did you apply the patch? I tested it as giving exactly that output after the patch.

Which patch? I just changed`tin.Witness != nil` to `len(tin.Witness) != 0` ? But that shouldn't effect the result, just prevent it erroring out. (The reason I didn't hit it before was I was loading the transaction in a diff way that had actual nil witness)

AdamISZ:
Ah, I get it now.
The Serialize() call is failing so the Raw is not valid, so the fact that you get a weird result instead of an error is related to your point about Raw (that is, it being an error prone way of doing things).

I've basically finished all the changes (including that), am about to push. It'll be fixed then.

AdamISZ:
So the above commit does those three things (fix validateUnsignedTx, remove Raw and use Serialize() only, change Inputs, Outputs and Unknowns in Psbt structs to be slices not pointers-to (arguably could be slices of pointers, but this seems fine).

Need to improve the tests probably, also create a decode (into json say) function seems like an obvious win. Exact nature of rpcs and where they go is open (as mentioned at the start, did a little bit of work on that in my lnd fork in lnwallet, but that's just speculative really).

AdamISZ:
@RHavar Just in case it helps, when I said it tested correct, I meant [this function](https://github.com/btcsuite/btcutil/pull/126/commits/2d9506364c3a59c9e13657aadda0bd70b397c3b1#diff-ee80c36ccb127a2fe07076b7496503b4R1022) which afaict is exactly the same calls as yours. It worked after that change (len(tin.Witness)). A bit confused tbh what was happening before but if it's still not operating correctly let me know.

AdamISZ:
Just dropping a note here that [this](https://github.com/bitcoin/bitcoin/pull/14197) update to functionality that's been done in Core must be added here, too.

psbt/psbt.go
```diff
@@ -0,0 +1,957 @@
...
+			}
+			pi.WitnessUtxo = txout
+
+		case PsbtInPartialSig:
+			newPartialSig := PartialSig{PubKey: keydata,
+				Signature: value}
+			if !newPartialSig.checkValid() {
+				return ErrInvalidPsbtFormat
+			}
+			pi.PartialSigs = append(pi.PartialSigs, &newPartialSig)
```
afk11: https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#handling-duplicated-keys

I think this should check that the public key hasn't already added a signature

psbt/psbt.go
```diff
@@ -0,0 +1,957 @@
...
+
+		case PsbtInBip32Derivation:
+			if !validatePubkey(keydata) {
+				return ErrInvalidPsbtFormat
+			}
+			master, derivationPath, err := readBip32Derivation(value)
+			if err != nil {
+				return err
+			}
+			pi.Bip32Derivation = append(pi.Bip32Derivation,
```
afk11: Unique keys might also need checking here

psbt/psbt.go
```diff
@@ -0,0 +1,957 @@
...
+			pi.FinalScriptWitness = value
+
+		default:
+			keyintanddata := []byte{byte(keyint)}
+			keyintanddata = append(keyintanddata, keydata...)
+			newUnknown := &Unknown{
+				Key:   keyintanddata,
+				Value: value,
+			}
+			pi.Unknowns = append(pi.Unknowns, newUnknown)
```
afk11: these need a unique key also

psbt/psbt.go
```diff
@@ -0,0 +1,957 @@
...
+
+		case PsbtOutBip32Derivation:
+			if !validatePubkey(keydata) {
+				return ErrInvalidKeydata
+			}
+			master, derivationPath, err := readBip32Derivation(value)
+			if err != nil {
+				return err
+			}
+			po.Bip32Derivation = append(po.Bip32Derivation,
```
afk11: this should be the last of these, but unique keys here

AdamISZ:
Just to note on the above, I'd have preferred to do it a cleaner way involving having each type reuse the same code for duplicate checking, but I found it was a bit weird trying to do that with an interface; if there's a better idiom for it to remove that code duplication, let me know; what is here should work fine I guess, plus I added two tests for the deserialization of such cases.

psbt/finalizer.go
```diff
@@ -0,0 +1,452 @@
...
+		expectedSighashType = input.SighashType
+	}
+
+	return expectedSighashType == txscript.SigHashType(sig[len(sig)-1])
+}
+
+// isFinalized considers this input finalized if it contains
+// at least one of the FinalScriptSig or FinalScriptWitness
+// are filled (which only occurs in a successful call to Finalize*)
+func isFinalized(p *Psbt, idx int) bool {
```
bjarnemagnussen: Maybe (and very pedantic): Sometimes you use `idx` in argument to denote an input index, and other times `inIndex`. Just for consistency maybe choose one.

psbt/creator.go
```diff
@@ -0,0 +1,70 @@
...
+	nSequences []uint32) error {
+	// Create the new struct; the input and output lists will be empty,
+	// the unsignedTx object must be constructed and serialized,
+	// and that serialization should be entered as the only entry for
+	// the globalKVPairs list.
+
+	// Check the version is a valid Bitcoin tx version; the nLockTime
+	// can be any valid uint32. There must be one sequence number per
+	// input.
+	if !(Version == 1 || Version == 2) || len(nSequences) != len(inputs) {
```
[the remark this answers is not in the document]
AdamISZ: I agree. I had the same issue when writing that, I think that if btcd gets updated, then this can, later.

psbt/finalizer.go
```diff
@@ -0,0 +1,452 @@
...
+		positionMap = append(positionMap, kv{Key: pos, Value: p})
+	}
+
+	sort.Slice(positionMap, func(i, j int) bool {
+		return positionMap[i].Key < positionMap[j].Key
+	})
+	// Build the return array of signatures
+	sigsNew := [][]byte{}
+	for _, x := range positionMap {
+		sigsNew = append(sigsNew, x.Value[1])
```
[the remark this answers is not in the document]
AdamISZ: It does seem like a really tiny difference.

psbt/finalizer.go
```diff
@@ -0,0 +1,452 @@
...
+	if len(sigs) < 1 || len(pubKeys) < 1 {
+		// We have failed to identify at least 1 (sig, pub) pair
+		// in the PSBT, which indicates it was not ready to be finalized.
+		return ErrNotFinalizable
+	}
+
+	if !containsRedeemScript {
+		// p2pkh
+		builder := txscript.NewScriptBuilder()
+		builder.AddData(sigs[0]).AddData(pubKeys[0])
```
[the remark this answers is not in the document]
AdamISZ: Yes I agree, I will add a check here. P2PKH is by definition single-key.
This comment would also apply at line 426.

psbt/finalizer.go
```diff
@@ -0,0 +1,452 @@
...
+		if err != nil {
+			return err
+		}
+	} else {
+		// This is assumed p2sh multisig
+		// Given redeemScript and pubKeys we can decide in what order
+		// signatures must be appended.
+		orderedSigs, err := extractKeyOrderFromScript(pInput.RedeemScript,
+			pubKeys, sigs)
+		if err != nil {
```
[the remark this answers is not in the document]
AdamISZ: Yes, I like/agree with your points here. Obviously only supporting multisig is a bit crappy, I think perhaps refactor a bit so at least there is a "not implemented" function to support generic Script rather than msig only, which can be added later, and also your point about checking by popping off the stack makes a lot of sense. I will try to do that part at least.

psbt/finalizer.go
```diff
@@ -0,0 +1,452 @@
...
+		} else {
+			// Otherwise, we must have a witnessScript field,
+			// to fulfil the requirements of p2wsh
+			// NOTE (we tacitly assume multisig)
+			if !cointainsWitnessScript {
+				return ErrNotFinalizable
+			}
+			witness, err = getMultisigScriptWitness(pInput.WitnessScript,
+				pubKeys, sigs)
+			if err != nil {
```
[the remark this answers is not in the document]
AdamISZ: Yes

psbt/finalizer.go
```diff
@@ -0,0 +1,452 @@
...
+		builder := txscript.NewScriptBuilder()
+		builder.AddData(pInput.RedeemScript)
+		scriptSig, err = builder.Script()
+		if err != nil {
+			return err
+		}
+		if !cointainsWitnessScript {
+			// Assumed p2sh-p2wkh
+			// Here the witness is just (sig, pub)
+			witness, err = writePKHWitness(sigs[0], pubKeys[0])
```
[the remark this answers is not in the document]
AdamISZ: Yep, as above, agreed.

psbt/finalizer.go
```diff
@@ -0,0 +1,452 @@
...
+			if err != nil {
+				return err
+			}
+		} else {
+			// Assumed p2sh-p2wsh with multisig.
+			// To build the witness, we do exactly as for the native p2wsh case.
+			witness, err = getMultisigScriptWitness(pInput.WitnessScript,
+				pubKeys, sigs)
+			if err != nil {
```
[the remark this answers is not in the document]
AdamISZ: Yes

psbt/finalizer.go
```diff
@@ -0,0 +1,452 @@
...
+package psbt
+
+// The Finalizer requires provision of a single PSBT input
+// in which all necessary signatures are encoded, and
+// uses it to construct valid final scriptSig and scriptWitness
+// fields.
+// NOTE that Finalizer is an interface, since it's understood
+// that the logic to finalize inputs varies per input type;
+// in this module are implementations for native segwit p2sh 2 of 2 multisig
+// and non-segwit p2sh 2 of 2 multisig.
```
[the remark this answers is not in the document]
AdamISZ: Yes this is a stale comment. Will fix.

psbt/finalizer.go
AdamISZ: OK, will fix.

AdamISZ:
@bjarnemagnussen many thanks for your review. I'm sorry I have left this to one side for a few months, was busy with other stuff. I also notice you're doing a python implementation which may be of interest to me for Joinmarket, but that's off topic here.

Made a few comment responses above, I will try to address those points and update the PR shortly.

psbt/finalizer.go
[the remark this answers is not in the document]
AdamISZ: Done in 1d79a3e2102b4c8b4929d0f446de18b65453f03f

psbt/finalizer.go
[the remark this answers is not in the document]
AdamISZ: Partially done in 1d79a3e2102b4c8b4929d0f446de18b65453f03f - will next look at the M count (ideally parse it properly but not clear if what's needed is exported, we'll see)

psbt/finalizer.go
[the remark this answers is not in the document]
AdamISZ: Done in 1d79a3e2102b4c8b4929d0f446de18b65453f03f

psbt/finalizer.go
AdamISZ: Done in 1d79a3e2102b4c8b4929d0f446de18b65453f03f

psbt/finalizer.go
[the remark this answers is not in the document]
AdamISZ: Done in 1d79a3e2102b4c8b4929d0f446de18b65453f03f

psbt/finalizer.go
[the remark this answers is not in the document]
AdamISZ: Done in 021250ec3443533a7e9139f7ad1580488489bcf3 - see comment below.

psbt/finalizer.go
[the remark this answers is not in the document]
AdamISZ: Done in 021250ec3443533a7e9139f7ad1580488489bcf3 see comment below.

psbt/finalizer.go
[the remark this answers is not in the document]
AdamISZ: Done in 021250ec3443533a7e9139f7ad1580488489bcf3 see comment below.

AdamISZ:
So https://github.com/btcsuite/btcutil/pull/126/commits/021250ec3443533a7e9139f7ad1580488489bcf3 adds checks that a given witnessScript or redeemScript is actually multisig and that the provided pubkeys and signatures in the partial signatures match, in number, the policy of the multisig.

This was achieved using `txscript.CalcMultiSigStats` in the `standard.go` module.

As mentioned above we're still restricted to redeemScripts/witnessScripts of type M of N multisig here, anything else will not be finalizable. It may need more work on my part (or others) to generalize that on this codebase. But at least we cover for now all of p2pkh, p2sh multisig, p2wpkh, p2sh-p2wpkh, p2sh-p2wsh (multisig).

I'll take a little look at linting and other code checks that are appropriate here, and probably add the new "POR" keytype that was added for inputs recently to BIP174, but after that there'll probably not be much more I can do to improve it.

psbt/creator.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// createPsbt, on provision of an input and output 'skeleton' for
+// the transaction, returns a Creator struct.
+// Note that we require OutPoints and not TxIn structs, as we will
+// only populate the txid:n information, *not* any scriptSig/witness
+// information. The values of nLockTime, nSequence (per input) and
+// transaction version (must be 1 of 2) must be specified here. Note
+// that the default nSequence value is wire.MaxTxInSequenceNum.
+func (c *Creator) createPsbt(inputs []*wire.OutPoint,
```
philipglazman: Shouldn't this method be exported so that code outside of package can create a psbt?

AdamISZ: Hmm, yes! Strange oversight; there's NewPsbtFromUnsignedTx and NewPsbt (which reads in a serialization instead of creating from scratch) exported in the main psbt.go but for some reason I overlooked that this method ought to be exported too. Come to think of it, the other two methods should perhaps be more logically in this file also.

AdamISZ: Done in cee0fa461626d53d87dc3f0b8880af05e9bfbd53

Roasbeef:
Has this PR been updated to account for these changes in `bitcoind`?

AdamISZ:
I remember around the last time I was looking at this, something like May, there was at least a new field added (POR as mentioned above), and I haven't looked at it since then.
So yes for full support, those additions would be needed (although with BIP174's design cross-compatibility should still be there even if other impls know about fields you don't). Also I'd have to review Core merges to know if there are further things to do.

PR #127, opened 2018-09-08
gcs/builder: Omit all OP_RETURN scripts from basic filter.

jimpo:


Roasbeef:
Agree on clarification in the BIP, I actually have a commit that spells it out. Will make PR to the BIP repo for that, thanks for this anyway!

PR #136, opened 2019-03-03
Fix block transactions range check error

imnisen:
Max transaction index should be `numTx-1` not `numTx`


jcvernaleo:
Done elsewhere.

PR #146, opened 2019-07-19
bloom: add merkle block verifier

wpaulino:
This allows us to verify the proof contained within a `merkleblock` and extract the transactions it commits to.


bjornoj:
I think it would be good to have tests testing the new `NewMerkleBlockWithTxs`

PR #150, opened 2019-12-11
support Go Modules

clarenous:
Is it a proper time to use Go Modules now?

All test passed.

In this pull request `github.com/btcsuite/btcd v0.20.1-beta` is required in `go.mod`.


jakesyl:
@jcvernaleo (as per btcsuite/btcd#1530)

- Low priority
- Outdated

Should be closed now

PR #151, opened 2019-12-18
add test case for litecoin bech32 addresses

michael1011:
I noticed that the Litecoin LND wallet on regtest, simnet, testnet and mainnet doesn't support sending to Litecoin bech32 addresses because the list of valid bech32 prefixes is hardcoded in `btcd` and, of course, doesn't include the Litecoin ones.

Therefore I am opening this PR that fixes sending to bech32 Litecoin addresses in the LND wallet by comparing the bech32 prefix to the one specified in the parameter `defaultNet *chaincfg.Params`.

### DIFF
--- a/address_test.go
+++ b/address_test.go
@@ -13,25 +13,32 @@ import (
 	"testing"
 
 	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 	"golang.org/x/crypto/ripemd160"
 )
 
 type CustomParamStruct struct {
+	Net              wire.BitcoinNet
 	PubKeyHashAddrID byte
 	ScriptHashAddrID byte
+	Bech32HRPSegwit  string
 }
 
 var CustomParams = CustomParamStruct{
-	PubKeyHashAddrID: 0x30, // starts with L
-	ScriptHashAddrID: 0x32, // starts with M
+	Net:              0xdbb6c0fb, // litecoin mainnet HD version bytes
+	PubKeyHashAddrID: 0x30,       // starts with L
+	ScriptHashAddrID: 0x32,       // starts with M
+	Bech32HRPSegwit:  "ltc",      // starts with ltc
 }
 
 // We use this function to be able to test functionality in DecodeAddress for
 // defaultNet addresses
 func applyCustomParams(params chaincfg.Params, customParams CustomParamStruct) chaincfg.Params {
+	params.Net = customParams.Net
 	params.PubKeyHashAddrID = customParams.PubKeyHashAddrID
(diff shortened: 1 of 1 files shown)


losh11:
It's a small fix and without this fix it's blocking litecoin lightning apps on our side so 👍

Roasbeef:
This should live in the `ltcsuite` repos instead.

michael1011:
> This should live in the ltcsuite repos instead.

@Roasbeef as far as I understand it, LND is always using `btcutil` as dependency and merging this fix in `ltcsuite/ltcutil` would have no effect on LND. But feel free to correct me if I am wrong.

address_test.go
```diff
@@ -20,18 +20,21 @@ import (
...
+	ScriptHashAddrID: 0x32,  // starts with M
+	Bech32HRPSegwit:  "ltc", // starts with ltc
 }
 
 // We use this function to be able to test functionality in DecodeAddress for
 // defaultNet addresses
 func applyCustomParams(params chaincfg.Params, customParams CustomParamStruct) chaincfg.Params {
 	params.PubKeyHashAddrID = customParams.PubKeyHashAddrID
 	params.ScriptHashAddrID = customParams.ScriptHashAddrID
+	params.Bech32HRPSegwit = customParams.Bech32HRPSegwit
```
onyb, line 37: This is just missing a call to `chaincfg.Register`. You need to do something like this:

```go
if err := chaincfg.Register(&params); err != nil {
        panic(err)
}
```

You may also need to add the following to ensure that there are no duplicate network parameters, since we copy the config from `chaincfg.MainNetParams`.

```golang
params.Net = 0xdeadbeef  // whatever the value is for Litecoin
```

michael1011:
@onyb your suggestion worked! I guess that the issue in lnd is related to not calling `chaincfg.Register()` when using a litecoin network then...

I removed my changes in `address.go` but updated and left the tests for the bech32 litecoin address.

address_test.go
```diff
@@ -13,25 +13,32 @@ import (
...
+	Net              wire.BitcoinNet
 	PubKeyHashAddrID byte
 	ScriptHashAddrID byte
+	Bech32HRPSegwit  string
 }
 
 var CustomParams = CustomParamStruct{
-	PubKeyHashAddrID: 0x30, // starts with L
-	ScriptHashAddrID: 0x32, // starts with M
+	Net:              0xdbb6c0fb,
```
onyb, line 29: Can you please add a comment here? Example below:
```go
Net:              0xdbb6c0fb,  // litecoin mainnet HD version bytes
```

Also, for my reference, can you please comment here the link where this is documented?

michael1011: @onyb I took the value from ltcd: https://github.com/ltcsuite/ltcd/blob/master/wire/protocol.go#L150

PR #156, opened 2020-02-25
Fix tiny bug in tx.go

zkvm:
Tiny bug fixed in tx.go, easy diff.


zkvm:
> Superceded by #165.

OK Thanks. Closed

PR #154, opened 2020-01-06
Fix sanity test and ordering in checkIsMultiSigScript

philipglazman:
Fixes return value ordering expected from txscript.CalcMultiSigStats.

Change sanity check to make sure that **m** of n pubkeys are provided.

PR #157, opened 2020-02-26
Assign returned values in correct order

kallerosenbaum:
This is basically a copy of PR #154, which is now obsolete due to refactorings. It also adds a test case for this issue.


jakesyl:
Looks good to me

philipglazman:
looks good!

PR #165, opened 2020-05-03
Fix nil pointer dereference when WitnessHash is called before HasWitness

onyb:
If the witness hash is populated in the cache before the `HasWitness` method is invoked, it results in a nil pointer deference.

Added a breaking test that reproduces the behaviour; then a few assertions using dummy witness data.

Issue: btcsuite/btcd#1543
Related PR: btcsuite/btcutil#156

### DIFF
--- a/tx_test.go
+++ b/tx_test.go
@@ -95,3 +95,42 @@ func TestTxErrors(t *testing.T) {
 			"got %v, want %v", err, io.EOF)
 	}
 }
+
+// TestTxHasWitness tests the HasWitness() method.
+func TestTxHasWitness(t *testing.T) {
+	msgTx := Block100000.Transactions[0] // contains witness data
+	tx := btcutil.NewTx(msgTx)
+
+	tx.WitnessHash() // Populate the witness hash cache
+	tx.HasWitness()  // Should not fail (see btcsuite/btcd#1543)
+
+	if !tx.HasWitness() {
+		t.Errorf("HasWitness: got false, want true")
+	}
+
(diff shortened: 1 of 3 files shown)


tx_test.go
```diff
@@ -95,3 +95,31 @@ func TestTxErrors(t *testing.T) {
...
+		txWithoutWitness := btcutil.NewTx(msgTxWithoutWitness)
+		if txWithoutWitness.HasWitness() {
+			t.Errorf("HasWitness: got false, want true")
+		}
+		if !txWithoutWitness.WitnessHash().IsEqual(txWithoutWitness.Hash()) {
+			t.Errorf("WitnessHash: witness hash and tx id must be same - "+
+				"got %v, want %v", txWithoutWitness.WitnessHash(), txWithoutWitness.Hash())
+		}
+	}
+}
```
Rjected, line 125: While it's great that you're testing both methods, it's usually more idiomatic to test each method independently. Then, in regression testing, if `TestTxHasWitness` breaks and a new test, `TestTxHashWitness`, does not, then we more or less know the bug is in `TestTxHasWitness` and can focus on that, rather than trying to figure out which method is incorrect after seeing that only `TestTxWitness` fails.

onyb: Done! Although, I wanted to include a proof-of-fix in my PR, so in the test for `HasWitness()`, I'm calling the `WitnessHash()` first, since that's how the bug is reproduced. Let me know if this is acceptable.

Rjected: I think that makes sense, looks good!

johnsonjh:
..

PR #168, opened 2020-05-17
Support decoding with length limit as a param

assafmo:
`Decode` can now accept char limit instead of using `90` as a constant.

This caused a bug on cosmos-sdk, for example, where they're sometimes using addresses with a length of 218 characters: https://github.com/cosmos/cosmos-sdk/pull/6237


Roasbeef:
Closing this as https://github.com/btcsuite/btcutil/pull/177 was merged.

PR #130, opened 2018-10-28
hdkeychain: extend API with ChainCode(), ChildNum()

benma:
No way to access them otherwise. Handy for interoperability.

PR #171, opened 2020-06-18
hdkeychain: expose extended key's version and child index

wpaulino:


Roasbeef:
With this #130 can be reduced to just adding the `ChainCode` method.

PR #173, opened 2020-06-19
psbt: add/export new utility functions and BIP 69 in-place sort

guggero:
To be able to implement PSBT funding in `btcwallet` we need some common utility functions added and some exported to be used externall.

We also add the ability to sort a PSBT according to [BIP 69](https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki). The difference to the normal wire TX sort is that the partial inputs and outputs that must match the wire transaction's `TxIn` and `TxOut` are sorted as well.

### DIFF
--- a/psbt/utils.go
+++ b/psbt/utils.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 	"sort"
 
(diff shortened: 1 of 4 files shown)


psbt/utils.go
```diff
@@ -15,10 +15,10 @@ import (
 	"github.com/btcsuite/btcd/wire"
 )
 
-// writeTxWitness is a A utility function due to non-exported witness
+// WriteTxWitness is a A utility function due to non-exported witness
```
wpaulino, line 19:
```suggestion
// WriteTxWitness is a utility function due to non-exported witness
```

PR #167, opened 2020-05-13
build: replace travis with github ci

dajohi:

PR #177, opened 2020-07-14
bech32: back port improvements from decred/dcrd

### LINKED ISSUE #152: bech32: Encode accepts invalid uppercase HRP
bech32.Encode will accept an uppercase HRP and generate an invalid bech32 encoding instead of normalizing it or returning an error.

BIP 0173 is clear that mixed case encodings are invalid, and that the lowercase encoding should be used for checksum purposes. This means there are two ways to handle an uppercase HRP in Encode:

1. treat it as lowercase, and ideally return a lowercase encoding
2. return an error

Instead, this library will use the uppercase values for the checksum and return a mixed case encoding. Decode will correctly reject the mixed case encoding, and will fail the checksum of the lowercase version.

https://play.golang.org/p/h0ekj8VmiPV

onyb:
### What is this about?

This PR brings a host of improvements to the bech32 package, back porting from the following commits:
* decred/dcrd@9b88dd0
* decred/dcrd@3a0dce1
* decred/dcrd@3cfb088

**The public interface of the package remains unchanged**, of course not including new public methods.

### Summary of changes

* Improved error handling using dedicated error types. Programmatically detect if the errors produced are the expected ones.
* Improve test coverage to test more corner cases. Added test vectors from Bitcoin Core.
* Add a benchmark for a full encode/decode cycle of a bech32 string.
* Add a new function `DecodeNoLimit`, for decoding large bech32 encoded strings. It does NOT validate against the BIP-173 maximum length allowed for bech32 strings.
* Automatically convert the HRP to lowercase in `Encode` function.
* Improve performance of `Encode`/`Decode` functions by using `strings.Builder`.
* Improve memory allocation in `ConvertBits` function.
* Updated documentation.

Credits: @matheusd

Closes #152 and #168.

### DIFF
--- a/bech32/bech32.go
+++ b/bech32/bech32.go
@@ -150,12 +150,13 @@ func bech32VerifyChecksum(hrp string, data []byte) bool {
 }
 
 // DecodeNoLimit decodes a bech32 encoded string, returning the human-readable
-// part and the data part excluding the checksum. This function does NOT
+// part and the data part excluding the checksum.  This function does NOT
 // validate against the BIP-173 maximum length allowed for bech32 strings and
 // is meant for use in custom applications (such as lightning network payment
 // requests), NOT on-chain addresses.
 //
-// Note that the returned data is 5-bit (base32) encoded.
+// Note that the returned data is 5-bit (base32) encoded and the human-readable
+// part will be lowercase.
 func DecodeNoLimit(bech string) (string, []byte, error) {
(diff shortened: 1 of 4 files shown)


onyb:
Oops, Travis CI was testing against really old versions of Go. Additionally, `gometalinter` seems dead, and I don't really want to fix all the CI stuff in this PR. Let's have #167 merged first.

jcvernaleo (approved):
OK

PR #178, opened 2020-07-20
psbt: restore compatibility with wallets that patch CVE-2020-14199

guggero:
Fixes https://github.com/lightningnetwork/lnd/issues/4400.

Wallets that include the fix for the CVE-2020-14199 vulnerability always include the full non-witness UTXO even if the output being spent is a witness output.
To restore compatibility with those wallets, [Bitcoin Core removed the sanity checks on partial inputs](https://github.com/bitcoin/bitcoin/pull/19215).
This PR does the same for the `btcutil/psbt` library.

### DIFF
--- a/psbt/updater.go
+++ b/psbt/updater.go
@@ -102,6 +102,11 @@ func (p *Updater) addPartialSignature(inIndex int, sig []byte,
 		}
 	}
 
+	// Attaching signature without utxo field is not allowed.
+	if pInput.WitnessUtxo == nil && pInput.NonWitnessUtxo == nil {
+		return ErrInvalidPsbtFormat
+	}
+
 	// Next, we perform a series of additional sanity checks.
 	if pInput.NonWitnessUtxo != nil {
 		if len(p.Upsbt.UnsignedTx.TxIn) < inIndex+1 {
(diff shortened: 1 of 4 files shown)


Roasbeef:
cc @halseth

psbt/updater.go
```diff
@@ -102,6 +102,11 @@ func (p *Updater) addPartialSignature(inIndex int, sig []byte,
 		}
 	}
 
+	// Attaching signature without utxo field is not allowed.
```
Roasbeef, line 105: So this now forces all implementations to implement the check? I guess from a conservative PoV, this is the safest thing to do. However IMO, the bug itself was a bit overblown, in that it requires the user to sign multiple times (it's a fault attack essentially), and hardware wallets generally don't keep any state between signing attempts. 

Zooming out, does this mean that in order to implement the generic PSBT calls in the `lnd` side, we now require a txindex for the target backend? Also this means that nodes using the neutrino backend wouldn't be able to use it as it (this library at the very least) can't populate this data?

guggero: This is just the `else` case that was moved up. We even loosen some restrictions by allowing both UTXO fields to be set and checking both.

I also think the bug was overblown a bit. But I can see why a bug "that can cause you to lose money" sounds very bad for a HW wallet manufacturer.
But because most HW wallets are now patched, we need to at least be compatible with their PSBT packets and not reject them.

I don't think this change forces chain backends to enable the `txindex`. Because as an updater you only need to know about the UTXOs involved. And because wallets keep the full transaction records for UTXOs in a separate DB anyway (or at least `btcwallet` does), no TX lookup should be necessary.

We use the `Wallet.FetchInputInfo` to add the UTXO information which uses `UnstableAPI(w).TxDetails(txid)`. From what I can see this only does DB lookups and no queries to the backend.

Issue #172, opened 2020-06-18
why not add leading zero bytes to make childkey 32 bytes long?


bibibong:
See [extendedkey.go:297](https://github.com/btcsuite/btcutil/blob/b2bf7f89d674a3702182b7e15f52807896051af3/hdkeychain/extendedkey.go#L297)
The bytes length may be smaller than 32. Is there  any bug when deriving hardened private key?
The derived key with padding differs.

onyb:
We only ever need to pad the private key bytes when further deriving a child extended key. This is already taken care of here:
https://github.com/btcsuite/btcutil/blob/4649e4b73b34090724fff8a302751260c20551fd/hdkeychain/extendedkey.go#L250-L258

The other thing we care about is whether the integer corresponding to the private child key bytes is valid for the secp256k1 elliptic curve. Here it doesn't matter if the serialization was done in 32 bytes or less, as long as the underlying crypto library can convert it to a [**`big.Int`**](https://golang.org/pkg/math/big/#Int) number.

Note that the `hdkeychain` package is tested against the official BIP-0032 test vectors, so I would be very surprised if there is a bug of the nature you describe.

Let me know if this answers your question. Otherwise, if you can provide a failing test case, I'll gladly take a look.

silencer-Tsai:
From my point of view, this implementatin might get a wrong result in some case. As the [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#private-parent-key--private-child-key) said, in the case of computing a child extended hardened private key from the parent private key, I = HMAC-SHA512(Key = cpar, Data = 0x00 || ser256(kpar) || ser32(i)). 
As for implementation, see [extendedkey.go](https://github.com/btcsuite/btcutil/blob/4649e4b73b34090724fff8a302751260c20551fd/hdkeychain/extendedkey.go#L250-L272).
If the length of *k.key* is less than 32, for example 31, then the Data = 0x00 || 31 bytes || 0x00 || ser32(i). *ser256(kpar)* just performs like another key leading with this 31 bytes and followed one 0x00. This is not consistent with the regulations of BIP32.

Here is a test vector. 

Seed(Hex): 1cae71ac5ed584ff88a078a119512d12bb61e5398521785e123b6d08809d44b2
* Chain m
[...]

PR #182, opened 2020-09-01
Correct BIP-32 derivation issue

devrandom:
Fixes issue #172.

This issue affects 1 in 256 hardened key derivations.  For standard BIP-44 accounts, this affects the coin and account derivations, so 1 in 128 accounts will not be compatible with other BIP-44 implementations.

The intention is that wallet implementations using btcutil can use the `IsAffectedByIssue172` call to check if they are affected, and potentially migrate (or ask the user to create a new wallet).

### DIFF
--- a/hdkeychain/extendedkey.go
+++ b/hdkeychain/extendedkey.go
@@ -120,7 +120,7 @@ type ExtendedKey struct {
 // fields.  No error checking is performed here as it's only intended to be a
 // convenience method used to create a populated struct. This function should
 // only be used by applications that need to create custom ExtendedKeys. All
-// other applications should just use NewMaster, Child, or Neuter.
+// other applications should just use NewMaster, Derive, or Neuter.
 func NewExtendedKey(version, key, chainCode, parentFP []byte, depth uint8,
 	childNum uint32, isPrivate bool) *ExtendedKey {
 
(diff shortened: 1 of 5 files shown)


silencer-Tsai:
Well done!

hdkeychain/extendedkey.go
```diff
@@ -254,7 +254,9 @@ func (k *ExtendedKey) Child(i uint32) (*ExtendedKey, error) {
 		// When the child is a hardened child, the key is known to be a
 		// private key due to the above early return.  Pad it with a
 		// leading zero as required by [BIP32] for deriving the child.
-		copy(data[1:], k.key)
+		// Additionally, right align it if it's shorter than 32 bytes.
+		offset := 33 - len(k.key)
+		copy(data[offset:], k.key)
```
Roasbeef, line 259: Rather than silently changing the behavior of this method, which can cause issues for any wallet that relies on this quirk as is (say when rescanning), I think we should instead _break_ this method, and force callers to choose one or the other. 

In other words: add a new variable to this method, and/or a helper func that assumes a new default value for this variable.

hdkeychain/extendedkey_test.go
```diff
@@ -1088,3 +1089,42 @@ func TestMaximumDepth(t *testing.T) {
 		t.Fatal("Child: deriving 256th key should not succeed")
 	}
 }
+
+func TestLeadingZero(t *testing.T) {
```
Roasbeef, line 1093: Missing `godoc` comment.

hdkeychain/extendedkey_test.go
```diff
@@ -1088,3 +1089,42 @@ func TestMaximumDepth(t *testing.T) {
...
+	ii := 399
```
Roasbeef, line 1094: Where's this value ceom from?

Roasbeef (changes requested):
Thanks for this set of changes! Picking this over the older one assuming that we can get this one in sooner. Also the commit message should have the package/folder prefix as specified in the contribution guidelines.

hdkeychain/extendedkey_test.go
devrandom: added an explanation in the code

hdkeychain/extendedkey_test.go
devrandom: added

hdkeychain/extendedkey.go
devrandom: renamed `ChildNonStandard` to `DeriveNonStandard` and `Child` to `Derive`.  added a usage note

Roasbeef (approved):
LGTM 🏆

Roasbeef:
cc @guggero

guggero (approved):
LGTM!

PR #183, opened 2020-09-10
base58: improve encode and decode performance

egonelbre:
Individual operations on big.Int values are slow, hence do computations up to `58^10` using an uint64. This makes the code roughly 10x faster.


calebcase:
Anything we can do to get this reviewed?

onyb (commented):
@egonelbre Can you please add some notes about the algorithm in the comments of the `Decode` and `Encode` functions?

egonelbre:
@onyb hopefully, that's sufficient.

egonelbre:
@onyb ping, is there anything else that needs to be improved?

onyb:
@egonelbre Thanks for the reminder. I got busy with reviewing some of btcd PRs. Will take a deeper look today or tomorrow, and most likely have it merged on Tuesday (during our weekly sync meeting).

base58/base58.go
```diff
@@ -10,24 +10,63 @@ import (
...
+		answer.Mul(answer, bigRadix[n])
+
+		total := uint64(0)
+		for i := 0; i < n; i++ {
+			tmp := b58[t[i]]
+			if tmp == 255 {
+				return []byte("")
+			}
+			total = total*58 + uint64(tmp)
+		}
```
onyb, line 65: That's quite neat. The algorithm doesn't need to loop over the string in reverse order (unlike the previous version). It's similar to the way `ParseInt` works in most languages.

onyb (approved):
Great work on this PR. Looks really good! Thanks to the code comments, I was able to understand the changes better.

base58/base58.go
```diff
@@ -10,24 +10,63 @@ import (
...
+	// Of course, in addition, we'll need to handle boundary condition when `b` is not multiple of 58^10.
+	// In that case we'll use the bigRadix[n] lookup for the appropriate power.
+	for t := b; len(t) > 0; {
+		n := len(t)
+		if n > 10 {
+			n = 10
 		}
-		scratch.SetInt64(int64(tmp))
-		scratch.Mul(j, scratch)
+		answer.Mul(answer, bigRadix[n])
```
onyb, line 56: small nit: I was wondering if this line could be after the loop.

egonelbre: Not really, because it needs to be run after every 10 input bytes.

egonelbre: Sure, I'll move it there.

egonelbre: Fixed.

onyb:
One last thing  - could you please pull in the latest test cases from Bitcoin Core? Here's the link: https://github.com/bitcoin/bitcoin/blob/master/src/test/data/base58_encode_decode.json

egonelbre:
@onyb ping.

jcvernaleo (approved):
Looks good to me.  Will wait until the meeting tomorrow to merge in case we can get any more eyes on it, but nice work!

jakesyl (approved):

jcvernaleo:
Just going to add that most of the time we prefer to squash to single commits but in this case I think the three commits make very logic sense so leaving them un-squashed is good.

And thanks!

PR #185, opened 2020-09-25
allocate less in Encode

josharian:


onyb:
@josharian Needs a rebase following #183.

josharian:
If I take the time to rebase will someone review and merge? It has been sitting unreviewed for two months.

jcvernaleo:
Hi @josharian

Sorry for letting this one sit.  We are getting better at prioritizing and not missing PRs but aren't perfect at it yet so sometimes ones get missed.  As our backlog gets smaller I'm hoping for it to happen less though.  At least that is my hope :)

Anyway, yes, please rebase and I'll make sure we review it.  Will add it to our project board to make sure we don't miss it again.

josharian:
Done. #183 picked up the lowest hanging fruit that this PR addressed, but there was a still a small performance benefit available.

base58/base58.go
```diff
@@ -90,9 +89,11 @@ func Encode(b []byte) string {
 	x := new(big.Int)
 	x.SetBytes(b)
 
-	answer := make([]byte, 0, len(b)*136/100)
+	// maximum length of output is log58(2^(8*len(b))) == len(b) * 8 / log(58)
+	maxlen := int(float64(len(b))*1.365658237309761) + 1
```
onyb, line 93: You could probably get away with a `float32` here, which is still significantly bigger than what `len(b)` could ever be. Thoughts?

onyb (approved):
LGTM! Made a small comment (not that important).

base58/base58.go
josharian: My 2c: Doesn't really matter. There'll never be a detectable performance impact, it's equally readable either way. Not worth changing.

jcvernaleo (approved):
Looks good.
OK

PR #187, opened 2020-10-13
DecodeAddress: don't panic on nil defaultNet

matthewleon:
Commit 3ac1210 breaks `DecodeAddress`'s handling of `nil` `defaultNet`: it will panic. The documentation suggests that `DecodeAddress` shouldn't require a `defaultNet`, and indeed in the case of most Bitcoin addresses, with some small adjustments, it can work without one.

### DIFF
--- a/address.go
+++ b/address.go
@@ -171,6 +171,9 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
 	// Serialized public keys are either 65 bytes (130 hex chars) if
 	// uncompressed/hybrid or 33 bytes (66 hex chars) if compressed.
 	if len(addr) == 130 || len(addr) == 66 {
+		if defaultNet == nil {
+			return nil, errors.New("decoded address appears to be P2PK, but no network was provided")
+		}
 		serializedPubKey, err := hex.DecodeString(addr)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


address.go
```diff
@@ -186,24 +189,32 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
...
-		switch hash160 := decoded; {
-		case isP2PKH && isP2SH:
+
+	if len(decoded) == ripemd160.Size {
+		return decodeLegacyAddress(decoded, netID, defaultNet)
+	}
+	return nil, errors.New("decoded address is of unknown size")
+}
+
+func decodeLegacyAddress(hash160 []byte, netID byte, defaultNet *chaincfg.Params) (Address, error) {
```
Roasbeef, line 199: All lines should be wrapped to 80 character columns.

Roasbeef (changes requested):
Where does the documentation state that it's ok to pass a `nil` default net? The default itself is meant to be a fallback if we can't parse the address itself. it's also needed in certain cases to be able to distinguish even _what type_ of address is passed.

address.go
```diff
@@ -186,24 +189,32 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
...
+	paramsToTry := []*chaincfg.Params{defaultNet}
+	if defaultNet == nil {
+		paramsToTry = []*chaincfg.Params{
+			&chaincfg.MainNetParams, &chaincfg.TestNet3Params, &chaincfg.RegressionNetParams, &chaincfg.SimNetParams}
```
Roasbeef, line 203: This essentially re-implements `chaincfg.IsPubKeyHashAddrID` and co.

matthewleon:
> Where does the documentation state that it's ok to pass a nil default net?

It doesn't. At least not explicitly. My reading of it suggested that `defaultNet` would only be used in the case of addresses that don't encode the network:

> // The bitcoin network the address is associated with is extracted if possible.
// When the address does not encode the network, such as in the case of a raw
// public key, the address will be associated with the passed defaultNet.

But that's pretty subjective, and possible lazy, reading on my part.

PR #192, opened 2020-12-24
Add helper functions to generate (nested) segwit addresses from extended key

### LINKED ISSUE #105: BIP49 (SegWit) Address Generation Support
After spending a few hours on this, it seems to me that something is missing so I can't generate proper BIP49 addresses. Unless I missed something. I use hdkeychain.NewKeyFromString to generate (derived) addresses. Works fine for BIP44. But now I'm trying to generate SegWit addresses.

It seems I need to be able to set the BIP32 Derivation path to: m/49'/1'/0'/0, or, well m/49' something at least as currently m/44' seems to be hard-coded somewhere. I have the proper "Account Extended Public Key", but since I can't change the "Purpose", the generated addresses ends up different from what I expect to get.

Anyone here that can help me out here?

Thanks!

setpill:
Resolves #105 

Credit to @bjarnemagnussen for writing the original code. I manually rebased his patch on top of latest master, fixed a typo and renamed the current ExtendedKey.Address function to be in line with the naming scheme from the patch. In place of the renamed function a simple wrapper function with a deprecation notice was added for backwards compatibility.

### DIFF
--- a/hdkeychain/doc.go
+++ b/hdkeychain/doc.go
@@ -24,8 +24,8 @@ with the IsPrivate function.
 Transaction Signing Keys and Payment Addresses
 
 In order to create and sign transactions, or provide others with addresses to
-send funds to, the underlying key and address material must be accessible.  This
-package provides the ECPubKey, ECPrivKey, and Address functions for this
+send funds to, the underlying key and address material must be accessible. This
+package provides the ECPubKey, ECPrivKey, and various Address functions for this
 purpose.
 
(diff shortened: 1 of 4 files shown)


Roasbeef:
IMO this should live along the rest of the code in `btcutil.Address`, the only extra hop needed is to obtain the raw public key (which now has new types as of the `v2` version of this module). Closing this for now, can be re-opened in the `btcd` repo.

PR #196, opened 2021-05-14
psbt: bounds check SumUtxoInputValues with NonWitness.TxOut indexing

Crypt-iQ:
Otherwise, a malformed packet would panic on this check since the NonWitness.TxOut field did not have the required TxOuts necessary.


Roasbeef (approved):
LGTM ☀️

PR #197, opened 2021-05-20
btcutil: update module to v2

### LINKED ISSUE #193: Requesting New Release (v1.0.2 seems outdated)
Using this package in my code gives this error:

`masterKey.Derive undefined (type *hdkeychain.ExtendedKey has no field or method Derive)`

Seems like go modules refer to https://pkg.go.dev/github.com/btcsuite/btcutil -> github.com/btcsuite/btcutil v1.0.2  which don't include the commits after Apr 14, 2020. Looks like I need to refer to the [commit: ](https://github.com/btcsuite/btcutil/commit/dde9e31e50c913f5fa56ff1e405dd442dd12e452) - hdkeychain: correct BIP-32 derivation issue on Oct 21, 2020 for my code to work properly.

 Any help on this matter would be appreciated.

Roasbeef:
In this commit, we update the top-level go module as well as all the
relevant imports to point to the new v2 version of the module. This
change is long overdue after a breaking change was made to the
[`hdkeychain` package](github.com/btcsuite/btcutil) to fix a BIP 32
implementation divergence. A follow up PR will also update the `psbt`
module as well as significant changes have been made to the API and
logic since it was first added.

Fixes #193


Roasbeef:
Running into a bit of a chicken and the egg here: 
```
--- FAIL: TestUseBlockHash (0.00s)
    builder_test.go:89: Address script build failed: unable to generate payment script for unsupported address type *btcutil.AddressScriptHash
```

`txscript` is still using the `v1` version of the module, and `PayToAddrScript` does a type assertion to know which script to generate. 

Ofc, we can't update `txscript` yet as the module doesn't really exist (v2, this PR switches things over but doesn't include the tag).... 

Short term workaround would be to skip the test for now, then work on switching everything over on the `btcd` end, then make a final update here (which includes making everything a sub-module?).

Roasbeef:
Replaced by https://github.com/btcsuite/btcd/pull/1785

PR #198, opened 2021-05-27
Fix infinite recursion when printing error

guggero:
Commit 24e673ae introduced a bug that cause the Error() function to
call itself recursively forever, causing a stack overflow. We explicitly
cast the error to its base type to avoid the recursion and add a small
test case that would've triggered the bug before.


halseth (approved):
LGTM, good catch 👍

onyb (approved):
Looks good. 👍

Issue #195, opened 2021-05-14
Add support for the bech32m encoding defined in BIP-350


federicobond:
Taproot is just around the corner with a new encoding format for addresses. I already have a working version of btcutil with Bech32m, but it breaks the signature of `bech32.Encode` and `bech32.EncodeFromBase256` with an additional encoding parameter and also returns an additional encoding value in the `bech32.Decode` and `bech32.DecodeFromBase256`, which is used to make sure that addresses use the correct encoding depending on the witness version.

If we can't break the current API, then we need to find a new name for these functions or create a new `bech32m` package. Any comments or suggestions welcome.

gcsfred2:
@federicobond I suggest new names for these functions.

benma:
@federicobond would you mind opening a PR?

I agree that a new package is preferable.

benma:
https://github.com/btcsuite/btcutil/pull/202

PR #202, opened 2021-07-17
bech32: add new EncodeM and DecodeGeneric functions for bech32

Roasbeef:
In this commit, we add two new package level functions: `EncodeM`, and
`DecodeGeneric`. The new encode method is intended to allow callers to
specify that they want to use the new bech32m checksum. This should be
used when encoding segwit addresses with version 1 and beyond. The new
`DecodeGeneric` function allows a caller to decode a bech32 and bech32m
string with a single function. A new return value is added which is the
version of the returned bech32 string, which allows callers to perform
additional segwit addr validation (v1+ should use bech32m etc).

We opted to add new functions rather than modifying the existing
functions to not cause a breaking API change, as most uses in the wild
can just use the existing functions, and only taproot related logic/code
needs to worry about the new methods.

A series of tests have been added to ensure that `DecodeGeneric`
extracts the proper bech version, and we've also adopted the bech32m
tests from BIP 350.

Fixes part of https://github.com/btcsuite/btcd/issues/1735. 

Fixes https://github.com/btcsuite/btcutil/issues/195

### DIFF
--- a/bech32/bech32.go
+++ b/bech32/bech32.go
@@ -123,8 +123,11 @@ func bech32Polymod(hrp string, values, checksum []byte) int {
 // and 126), otherwise the results are undefined.
 //
 // For more details on the checksum calculation, please refer to BIP 173.
-func writeBech32Checksum(hrp string, data []byte, bldr *strings.Builder) {
-	polymod := bech32Polymod(hrp, data, nil) ^ 1
+func writeBech32Checksum(hrp string, data []byte, bldr *strings.Builder,
+	version Version) {
+
+	bech32Const := int(VersionToConsts[version])
+	polymod := bech32Polymod(hrp, data, nil) ^ bech32Const
 	for i := 0; i < 6; i++ {
 		b := byte((polymod >> uint(5*(5-i))) & 31)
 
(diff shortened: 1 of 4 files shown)


onyb:
Mentioning here so we don't forget - we should probably update the relevant Bitcoin wiki article: https://en.bitcoin.it/wiki/Bech32_adoption#Libraries

bech32/bech32.go
```diff
@@ -137,46 +140,55 @@ func writeBech32Checksum(hrp string, data []byte, bldr *strings.Builder) {
 
 // bech32VerifyChecksum verifies whether the bech32 string specified by the
 // provided hrp and payload data (encoded as 5 bits per element byte slice) has
-// the correct checksum suffix.
+// the correct checksum suffix. The version of bech32 used (bech32 OG, or
+// bech32m) is also retruend so allow the caller to perform proper address
```
alexbosworth, line 144:
```suggestion
// bech32m) is also returned so allow the caller to perform proper address
```

alexbosworth, line 144:
```suggestion
// bech32m) is also retruend to allow the caller to perform proper address
```

Roasbeef:
Fixed typos, fixed linter error.

bech32/bech32.go
```diff
@@ -202,34 +214,56 @@ func DecodeNoLimit(bech string) (string, []byte, error) {
...
 		// error contains this information.
 
 		// Extract the payload bytes and actual checksum in the string.
 		actual := bech[len(bech)-6:]
 		payload := decoded[:len(decoded)-6]
 
-		// Calculate the expected checksum, given the hrp and payload data.
+		// Calculate the expected checksum, given the hrp and payload
+		// data. We'll actually compute _both_ possibly valid checksum
+		// to further aide in debugging.
```
guggero, line 233: Cool, should be very useful to the user.

bech32/bech32_test.go
```diff
@@ -86,6 +87,109 @@ func TestBech32(t *testing.T) {
...
+		{"11llllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllludsr8", VersionM},
+		{"split1checkupstagehandshakeupstreamerranterredcaperredlc445v", VersionM},
+		{"?1v759aa", VersionM},
+
+		{"A12UEL5L", Version0},
+		{"a12uel5l", Version0},
+		{"an83characterlonghumanreadablepartthatcontainsthenumber1andtheexcludedcharactersbio1tt5tgs", Version0},
+		{"abcdef1qpzry9x8gf2tvdw0s3jn54khce6mua7lmqqqxw", Version0},
+		{"11qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqc8247j", Version0},
+		{"split1checkupstagehandshakeupstreamerranterredcaperred2y9e3w", Version0},
```
guggero, line 176: Should we throw a valid bech32 segwit v1 address in here just for completeness' sake? E.g. `bc1puxkz8vpy900c7z4q4302lkc3jjr2s42mayfqzzgr5yqdk5mgma3s0kntlh`

guggero (approved):
Great to have this implemented in a generic way. So we'd even be able to decode segwit v0 addresses that use bech32m (even if probably no wallet will ever create those).

LGTM :100:

bech32/bech32_test.go
Xekyo: Yeah, wouldn't hurt. I guess we'd recommend implementing all test vectors, the last in the valid section of the [v0-v16 test vectors](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki#test-vectors-for-v0-v16-native-segregated-witness-addresses), `bc1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vqzk5jj0`, also encodes a P2TR locking script.

Roasbeef:
> So we'd even be able to decode segwit v0 addresses that use bech32m (even if probably no wallet will ever create those).

Yeah exactly, the idea is that say in `lnd`, we can detect that case and reject the send w/ a detailed error message.

gitteri:
Looks like someone sent btc to the test vector mentioned above... most block explorers are unable to parse `bc1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vqzk5jj0`. Can we get this in soon? 
https://chain.so/address/BTC/bc1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vqzk5jj0

Roasbeef:
I've added all the additional test vectors. Note that this PR doesn't implement new versions of `btcuil.Address` for the new script versions, so some of the vectors don't apply as directly. I'll add them in the follow up PR that creates new `btcutil.Address` implementations for taproot addresses.

PR #204, opened 2021-09-15
Rename module cosmos/btcutil

conr2d:
The forked module is renamed `github.com/cosmos/btcutil`.

I tried to remove the original module dependency completely, but btcutil depends on `github.com/btcsuite/btcd` and btcd has type assertions with `github.com/btcsuite/btcutil`.

https://github.com/btcsuite/btcd/blob/master/txscript/standard.go#L440-L482

If we want to exclude the original module from dependencies, it's necessary to fork btcd, too. (not ideal)

Any other changes are simply replacing module name, but in `gcs/builder/builder_test.go`, btcutil is not this module, but the original one.


conr2d:
This PR is not for the upstream. Sorry for annoying.

jcvernaleo:
@conr2d no worries.  We've all done it at least once :)

Issue #200, opened 2021-07-10
Support single-key P2TR output scripts and addresses


benma:
- One more instance of the Address interface
- v1 witness scripts to be supported in [DecodeAddress()](https://github.com/btcsuite/btcutil/blob/e2ba6805a89060647fa22ec61b799acff6a4f6fd/address.go#L130)
- a new function named `NewAddressTaprootPubkey()` or similar.

More details: https://github.com/bitcoin/bips/blob/master/bip-0086.mediawiki

Related to bech32m support: https://github.com/btcsuite/btcutil/issues/195

benma:
Tracked also in https://github.com/btcsuite/btcd/issues/1735

Roasbeef:
This is next now that #195 is in

PR #205, opened 2021-09-24
address: add AddressTaproot

guggero:
Fixes https://github.com/btcsuite/btcutil/issues/200.

### DIFF
--- a/address.go
+++ b/address.go
@@ -506,25 +506,88 @@ func (a *AddressPubKey) PubKey() *btcec.PublicKey {
 	return a.pubKey
 }
 
+// AddressSegWit is the base address type for all SegWit addresses.
+type AddressSegWit struct {
+	hrp            string
+	witnessVersion byte
+	witnessProgram []byte
+}
+
+// EncodeAddress returns the bech32 (or bech32m for SegWit v1) string encoding
+// of an AddressSegWit.
+//
+// NOTE: This method is part of the Address interface.
+func (a *AddressSegWit) EncodeAddress() string {
+	str, err := encodeSegWitAddress(
+		a.hrp, a.witnessVersion, a.witnessProgram[:],
+	)
+	if err != nil {
+		return ""
+	}
+	return str
+}
+
+// ScriptAddress returns the witness program for this address.
+//
+// NOTE: This method is part of the Address interface.
+func (a *AddressSegWit) ScriptAddress() []byte {
+	return a.witnessProgram[:]
+}
+
+// IsForNet returns whether the AddressSegWit is associated with the passed
+// bitcoin network.
+//
+// NOTE: This method is part of the Address interface.
+func (a *AddressSegWit) IsForNet(net *chaincfg.Params) bool {
+	return a.hrp == net.Bech32HRPSegwit
+}
+
+// String returns a human-readable string for the AddressWitnessPubKeyHash.
+// This is equivalent to calling EncodeAddress, but is provided so the type
+// can be used as a fmt.Stringer.
+//
+// NOTE: This method is part of the Address interface.
+func (a *AddressSegWit) String() string {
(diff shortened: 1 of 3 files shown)


guggero:
Rebased and polished my old code I had lying around for this. PTAL @Roasbeef

address_test.go
```diff
@@ -655,13 +655,36 @@ func TestAddresses(t *testing.T) {
...
 		},
-		// Unsupported witness versions (version 0 only supported at this point)
+
+		// P2TR address tests.
 		{
-			name:  "segwit mainnet witness v1",
-			addr:  "bc1pw508d6qejxtdg4y5r3zarvary0c5xw7kw508d6qejxtdg4y5r3zarvary0c5xw7k7grplx",
-			valid: false,
-			net:   &chaincfg.MainNetParams,
+			name:    "segwit v1 mainnet p2tr",
```
Roasbeef, line 661: Let's also add these test vectors? https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki#test-vectors-for-v0-v16-native-segregated-witness-addresses

address_test.go
```diff
@@ -655,13 +655,36 @@ func TestAddresses(t *testing.T) {
 			},
 			net: &customParams,
 		},
-		// Unsupported witness versions (version 0 only supported at this point)
+
+		// P2TR address tests.
```
Roasbeef, line 659: I think we also want test vectors for the negative reject case: someone tries to use a segwit v1 witness program w/ the OG bech32 version/checksum

Roasbeef:
Once this is in, we'll also want to update this method as well: https://pkg.go.dev/github.com/btcsuite/btcd/txscript#PayToAddrScript 

As well as this type: https://pkg.go.dev/github.com/btcsuite/btcd/txscript#PkScript

address.go
```diff
@@ -149,8 +161,9 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
 			}
 
 			// We currently only support P2WPKH and P2WSH, which is
-			// witness version 0.
-			if witnessVer != 0 {
+			// witness version 0 and P2TR which is witness version
```
Roasbeef, line 164: We want to also have the `decodeSegWitAddress` method return the `bech32` version returned as well. This way we can reject decoding an address that's segwit v1, but that doesn't use the `bech32m` encoding.

address.go
```diff
@@ -681,3 +698,87 @@ func (a *AddressWitnessScriptHash) WitnessVersion() byte {
 func (a *AddressWitnessScriptHash) WitnessProgram() []byte {
 	return a.witnessProgram[:]
 }
+
+// AddressTaproot is an Address for a pay-to-taproot (P2TR) output. See BIP 341
+// for further details.
+type AddressTaproot struct {
+	hrp            string
```
Roasbeef, line 705: Extract this into a base `AddressSegwit` struct so we can save on the duplicaation here and above? This way it can also inherit all the shared methods used here.

address.go
Roasbeef: To accomplish this, we need to change the method to instead use `DecodeGeneric` so it can obtain the `bech32` version used when encoding.

Roasbeef (changes requested):

address_test.go
guggero: Yeah, added all of them. Also fixed some edge cases covered by those test vectors.

address_test.go
guggero: Right, this was made apparent by the test cases as well :+1:

address.go
guggero: Makes sense, was able to remove a lot of duplicated code!

ellemouton (approved):
LGTM 💥  Much excite!

PR #208, opened 2022-01-04
base58: fix bug with range

egonelbre:
The decode path was using a range over the string instead of iterating over the bytes.


Roasbeef:
Can you transfer this over to https://github.com/btcsuite/btcd? I can't move PRs the same way it's possible to move issues.

egonelbre:
Done https://github.com/btcsuite/btcd/pull/1799

